PR #2, opened 2018-11-23
CellML loading and expression printing

MichaelClerx:
Still a work in progress

### DIFF
--- a/fccodegen/__init__.py
+++ b/fccodegen/__init__.py
@@ -14,10 +14,13 @@ finally:
     # Always manually delete frame
     # https://docs.python.org/2/library/inspect.html#the-interpreter-stack
     del(frame)
+del(inspect)    # Don't expose imported modules
 
 # Template directory
 DIR_TEMPLATE = os.path.join(DIR_ROOT, 'templates')
-del(inspect)    # Don't expose as part of codegen
+
+# Directory for any other data
+DIR_DATA = os.path.join(DIR_ROOT, 'data')
 
 
 #
@@ -26,7 +29,7 @@ del(inspect)    # Don't expose as part of codegen
 with open(os.path.join(DIR_ROOT, 'version.txt'), 'r') as f:
     VERSION_INT = tuple([int(x) for x in f.read().split('.', 3)])
 VERSION = '.'.join([str(x) for x in VERSION_INT])
-del(os)
+del(os)     # Don't expose imported modules
 
 
 #
--- a/fccodegen/__init__.py
+++ b/fccodegen/__init__.py
@@ -52,6 +52,12 @@ def version(formatted=False):
 #
 from ._generator import (   # noqa
     create_weblab_model,
+    get_equations,
+    load_model,
     load_template,
+)
+
+from ._printer import (     # noqa
+    WebLabPrinter,
 )
 
--- a/fccodegen/_generator.py
+++ b/fccodegen/_generator.py
@@ -1,3 +1,7 @@
+#
+# Functions related to generating model code.
+# TODO: Find a better name/description/layout
+#
 import fccodegen
 import jinja2
 import os
@@ -15,13 +19,60 @@ def load_template(*name):
         return jinja2.Template(f.read())
 
 
+def load_model(path):
+    """
+    Loads a cellml model.
+    """
+    #TODO: This should become part of cellmlmanip
+
+    import cellmlmanip
+    import cellmlmanip.parser
+
+    model = cellmlmanip.parser.Parser(path).parse()
+    model.make_connections()
+    for c in model.components.values():
+        c.add_units_to_equations()
+    for e in model.equations:
+        model.check_left_right_units_equal(e)
+    return model
+
+
+def get_equations(model):
+    """
+    Gets the equations from a cellml model.
+    """
+    #TODO: This should become part of cellmlmanip
+
+    graph = model.get_equation_graph()
+
+    # Get sorted symbols
+    import networkx as nx
+    symbols = nx.lexicographical_topological_sort(graph, key=lambda x: str(x))
+
+    # Create list of equations
+    from sympy.physics.units import Quantity
+    eqs = []
+    for symbol in symbols:
+        # Get equation
+        eq = graph.nodes[symbol]['equation']
+
+        # Skip symbols that are not set with an equation
+        if eq is None:
+            continue
+
+        # Remove quantities (units)
+        eq = eq.subs({q: 1 for q in eq.atoms(Quantity)}, simultaneous=True)
+
+        eqs.append(eq)
+
+    return eqs
+
+
 def create_weblab_model(model, path):
     """
     Takes a :class:`cellmlmanip.Model`, generates a ``.pyx`` model for use
     with the Web Lab, and stores it at ``path``.
     """
-    from cellmlmanip import parser    
-    
     template = load_template('wl', 'weblab_model.pyx')
     with open(path, 'w') as f:
         f.write(template.render())
--- a/fccodegen/_generator.py
+++ b/fccodegen/_generator.py
@@ -20,6 +20,7 @@ def create_weblab_model(model, path):
     Takes a :class:`cellmlmanip.Model`, generates a ``.pyx`` model for use
     with the Web Lab, and stores it at ``path``.
     """
+    
     template = load_template('wl', 'weblab_model.pyx')
     with open(path, 'w') as f:
         f.write(template.render())
--- a/fccodegen/_generator.py
+++ b/fccodegen/_generator.py
@@ -20,6 +20,7 @@ def create_weblab_model(model, path):
     Takes a :class:`cellmlmanip.Model`, generates a ``.pyx`` model for use
     with the Web Lab, and stores it at ``path``.
     """
+    from cellmlmanip import parser    
     
     template = load_template('wl', 'weblab_model.pyx')
     with open(path, 'w') as f:
--- a/fccodegen/_printer.py
+++ b/fccodegen/_printer.py
@@ -1,0 +1,394 @@
+#
+# Methods to convert sympy expressions to Cython code.
+#
+# Should work with SymPy 1.1.1, but also later versions.
+#
+# Parts of this code are taken from:
+#  https://github.com/sympy/sympy/blob/master/sympy/printing/printer.py
+#  https://github.com/sympy/sympy/blob/master/sympy/printing/str.py
+#
+# Which comes with the following license:
+#
+# Copyright (c) 2006-2018 SymPy Development Team
+#
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are met:
+#
+#   a. Redistributions of source code must retain the above copyright notice,
+#      this list of conditions and the following disclaimer.
+#   b. Redistributions in binary form must reproduce the above copyright
+#      notice, this list of conditions and the following disclaimer in the
+#      documentation and/or other materials provided with the distribution.
+#   c. Neither the name of SymPy nor the names of its contributors
+#      may be used to endorse or promote products derived from this software
+#      without specific prior written permission.
+#
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+# AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+# IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+# ARE DISCLAIMED. IN NO EVENT SHALL THE REGENTS OR CONTRIBUTORS BE LIABLE FOR
+# ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+# DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+# SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+# CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+# LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
+# OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH
+# DAMAGE.
+#
+import sympy
+import sympy.printing
+
+from sympy.printing.precedence import precedence
+
+#from sympy.core import S, Rational, Pow, Basic, Mul
+#from .printer import Printer
+#from sympy.printing.precedence import PRECEDENCE
+#import mpmath.libmp as mlib
+#from mpmath.libmp import prec_to_dps
+#from sympy.utilities import default_sort_key
+
+# List of functions handled by python's `math` module
+_math_functions = {
+    'acos',
+    'acosh',
+    'asin',
+    'asinh',
+    'atan',
+    'atan2',
+    'atanh',
+    'ceil',     # called ceiling in sympy
+    'cos',
+    'cosh',
+    'exp',
+    'expm1',
+    'factorial',
+    'floor',
+    'log',
+    'log10',
+    'log1p',
+    'log2',
+    'sin',
+    'sinh',
+    'tan',
+    'tanh',
+}
+
+
+class WebLabPrinter(sympy.printing.printer.Printer):
+    """
+    Converts Sympy expressions to strings for use in Web Lab Cython model code.
+
+    To use, create a :class:`WebLabPrinter` instance, and call its method
+    :meth:`doprint()` with a Sympy expression argument.
+
+    Arguments:
+
+    ``symbol_function``
+        A function that converts symbols to strings (variable names).
+    ``derivative_function``
+        A function that converts derivatives to strings.
+
+    """
+
+    def __init__(self, symbol_function=None, derivative_function=None):
+        super(WebLabPrinter, self).__init__(None)
+
+        # Prefix for functions
+        self._prefix = 'math.'
+
+        # Symbol and derivative handling (default)
+        if symbol_function is None:
+            self._symbol_function = lambda x: str(x)
+        else:
+            self._symbol_function = symbol_function
+
+        if derivative_function is None:
+            self._derivative_function = lambda x: str(x)
+        else:
+            self._derivative_function = derivative_function
+
+    def _bracket(self, expr, parent_precedence):
+        """
+        Converts ``expr`` to string, and adds parentheses around the resut, if
+        and only if ``precedence(expr) < parent_precedence``.
+        """
+        if precedence(expr) < parent_precedence:
+            return '(' + self._print(expr) + ')'
+        return self._print(expr)
+
+    def _bracket_args(self, args, parent_precedence):
+        """
+        Applies :meth:`_bracket()` to a list of expressions, and joins them
+        with a comma.
+        """
+        return ', '.join([self._bracket(x, parent_precedence) for x in args])
+
+    def emptyPrinter(self, expr):
+        """
+        Called by :class:`Printer` as a last resort for unknown expressions.
+        """
+        raise RuntimeError(
+            'Unsupported expression type (' + str(type(expr)) + '): '
+            + str(expr))
+
+    def _print_Add(self, expr):
+        """ Handles addition & subtraction, with n terms. """
+        # This method is mostly copied from sympy.printing.Str
+
+        parts = []
+        my_prec = precedence(expr)
+        for term in expr.args:
+            # Don't use _bracket() here because we want to check the sign
+            t = self._print(term)
+
+            # Add sign
+            s = '+'
+            if t.startswith('-'):
+                s = '-'
+                t = t[1:]
+            parts.append(s)
+
+            # Add remaining term
+            parts.append('(' + t + ')' if precedence(term) < my_prec else t)
+
+        # Concatenate and return
+        if parts[0] == '+':
+            # Ignore leading plus
+            return ' '.join(parts[1:])
+        else:
+            # No space after first minus
+            return parts[0] + ' '.join(parts[1:])
+
+    def _print_And(self, expr):
+        """ Handles logical And. """
+        my_prec = precedence(expr)
+        return ' and '.join([self._bracket(x, my_prec) for x in expr.args])
+
+    def _print_BooleanFalse(self, expr):
+        """ Handles False """
+        return 'False'
+
+    def _print_BooleanTrue(self, expr):
+        """ Handles True """
+        return 'True'
+
+    def _print_Derivative(self, expr):
+        """ Handles Derivative objects. """
+        return self._derivative_function(expr)
+
+    def _print_Exp1(self, expr):
+        """ Handles the sympy E object """
+        return self._prefix + 'e'
+
+    def _print_float(self, expr):
+        """ Handles Python floats """
+        # Print short format if it doesn't change the value, else long format
+        short = str(expr)
+        if float(short) == expr:
+            return short
+        return '{: .17e}'.format(expr)
+
+    def _print_Float(self, expr):
+        """ Handles Sympy Float objects """
+        return self._print_float(float(expr))
+
+    def _print_Function(self, expr):
+        """ Handles function calls. """
+
+        # Check if function is known to python math
+        name = expr.func.__name__
+        if name == 'ceiling':
+            name = 'ceil'
+        if name not in _math_functions:
+            raise RuntimeError('Unsupported function: ' + str(name))
+
+        # Convert arguments and return
+        args = self._bracket_args(expr.args, 0)
+        return self._prefix + name + '(' + args + ')'
+
+    def _print_Infinity(self, expr):
+        return 'float(\'inf\')'
+
+    def _print_int(self, expr):
+        """ Handles python ints """
+        return str(expr)
+
+    def _print_Integer(self, expr):
+        """
+        Handles Sympy Integer objects, including special ones like Zero, One,
+        and NegativeOne.
+        """
+        return str(expr.p)
+
+    def _print_Mul(self, expr):
+        """
+        Handles multiplication & division, with n terms.
+
+        Division is specified as a power: ``x / y --> x * y**-1``.
+        Subtraction is specified as ``x - y --> x + (-1 * y)``.
+        """
+        # This method is mostly copied from sympy.printing.Str
+
+        # Check overal sign of multiplication
+        from sympy.core.mul import _keep_coeff
+        sign = ''
+        c, e = expr.as_coeff_Mul()
+        if c < 0:
+            expr = _keep_coeff(-c, e)
+            sign = '-'
+
+        # Collect all pows with more than one base element and exp = -1
+        pow_brackets = []
+
+        # Gather terms for numerator and denominator
+        a, b = [], []
+        for item in sympy.Mul.make_args(expr):
+
+            # Check if this is a negative power that we can write as a division
+            negative_power = (
+                item.is_commutative and item.is_Pow
+                and item.exp.is_Rational and item.exp.is_negative)
+            if negative_power:
+                if item.exp != -1:
+                    # Add as power
+                    b.append(sympy.Pow(item.base, -item.exp, evaluate=False))
+                else:
+                    # Add without power
+                    b.append(sympy.Pow(item.base, -item.exp))
+
+                    # Check if it's a negative power that needs brackets
+                    if (len(item.args[0].args) != 1
+                            and isinstance(item.base, sympy.Mul)):
+                        pow_brackets.append(item)
+
+            # Split Rationals over a and b, ignoring any 1s
+            elif item.is_Rational:
+                if item.p != 1:
+                    a.append(sympy.Rational(item.p))
+                if item.q != 1:
+                    b.append(sympy.Rational(item.q))
+
+            else:
+                a.append(item)
+
+        # Replace empty numerator with one
+        a = a or [sympy.S.One]
+
+        # Convert terms to code
+        my_prec = precedence(expr)
+        a_str = [self._bracket(x, my_prec) for x in a]
+        b_str = [self._bracket(x, my_prec) for x in b]
+
+        # Fix brackets for Pow with exp -1 with more than one Symbol
+        for item in pow_brackets:
+            if item.base in b:
+                b_str[b.index(item.base)] = \
+                    '(' + b_str[b.index(item.base)] + ')'
+
+        # Combine numerator and denomenator and return
+        a_str = sign + ' * '.join(a_str)
+        if len(b) == 0:
+            return a_str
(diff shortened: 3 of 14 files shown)


MichaelClerx:
Hi both! I'd appreciate any comments at this point

fccodegen/_generator.py
```diff
@@ -1,26 +1,180 @@
...
     #TODO: Look at the jinja PackageLoader. What's the advantage?
     with open(path, 'r') as f:
         return jinja2.Template(f.read())
 
 
+def load_model(path):
+    """
+    Loads a cellml model.
+    """
+    #TODO: This should become part of cellmlmanip
```
jonc125: Indeed :)

fccodegen/_generator.py
```diff
@@ -1,26 +1,180 @@
...
+
+
+def get_equations(graph):
+    """
+    Gets the equations from a cellml model graph.
+    """
+    #TODO: This should become part of cellmlmanip
+
+    # Get sorted symbols
+    #TODO: networkx is not a dependency
```
jonc125: As in, not listed as a dependency of `cellmlmanip`?

fccodegen/_generator.py
```diff
@@ -1,26 +1,180 @@
...
+        name = parts[-1]
+
+        # If already taken, rename _both_ variables using component name
+        if name in reverse:
+
+            # Get existing variable
+            other = reverse[name]
+
+            # Check it hasn't been renamed already
+            if symbols[other] == name:
```
jonc125: Shouldn't this always be true, since `symbols[reverse[name]] == name` always holds?

fccodegen/_generator.py
```diff
@@ -1,26 +1,180 @@
...
+
 def create_weblab_model(model, path):
     """
     Takes a :class:`cellmlmanip.Model`, generates a ``.pyx`` model for use
     with the Web Lab, and stores it at ``path``.
     """
+    # Get equation graph
+    graph = model.get_equation_graph()
+
+    # Get ordered equations
```
jonc125: Eventually when more API methods are in `cellmlmanip` this needs to happen below the definitions of `symbol_name` etc. and look more like:
```python
state_vars = model.get_state_variables()
free_var = model.get_free_variable()
derivatives = [sp.Derivative(state, free_var) for state in state_vars]
eqs = model.get_equations_for(derivatives)
```

fccodegen/_generator.py
```diff
@@ -1,26 +1,180 @@
...
+    # Create a list of state variable indices, variables, and derivative vars
+    states = []
+    for i, state in enumerate(get_state_list(graph)):
+        states.append((i, symbol_name(state), derivative_name(state)))
+
+    # Create the RHS equations
+    rhs_equations = []
+    for eq in eqs:
+        lhs, rhs = eq.lhs, eq.rhs
+        rhs_equations.append(p.doprint(lhs) + ' = ' + p.doprint(rhs))
```
jonc125: I'm kind of surprised that `p.doprint` can't take `eq` as an argument and do the right thing!

fccodegen/templates/wl/weblab_model.pyx
```diff
@@ -1,47 +1,50 @@
...
+    #TODO: What's this?
     model = <object>user_data
     cdef np.ndarray[Sundials.realtype, ndim=1] parameters = <np.ndarray>model.parameters
-    # State variables
-    cdef double var_rapid_time_dependent_potassium_current_Xr1_gate__Xr1 = (<Sundials.N_VectorContent_Serial>y.content).data[0]
-    cdef double var_rapid_time_dependent_potassium_current_Xr2_gate__Xr2 = (<Sundials.N_VectorContent_Serial>y.content).data[1]
 
+    # Unpack state variables
+    {%- for state in states %}
+    cdef double {{state[1]}} = (<Sundials.N_VectorContent_Serial>y.content).data[{{state[0]}}]
```
jonc125: It might be nicer to make `states` a list of dictionaries so you can do things like `state.var_name` or `state.index` here.

fccodegen/templates/wl/weblab_model.pyx
```diff
@@ -1,47 +1,50 @@
...
     # Mathematics
-    cdef double d_dt_rapid_time_dependent_potassium_current_Xr1_gate__Xr1 = ((1.0 / (1.0 + math.exp(( -26.0 - parameters[3]) * 0.14285714285714285))) - var_rapid_time_dependent_potassium_current_Xr1_gate__Xr1) / (1.0 * (450.0 / (1.0 + math.exp(( -45.0 - parameters[3]) * 0.10000000000000001))) * (6.0 / (1.0 + math.exp((parameters[3] + 30.0) * 0.086956521739130432)))) # per_millisecond
-    cdef double d_dt_rapid_time_dependent_potassium_current_Xr2_gate__Xr2 = ((1.0 / (1.0 + math.exp((parameters[3] + 88.0) * 0.041666666666666664))) - var_rapid_time_dependent_potassium_current_Xr2_gate__Xr2) / (1.0 * (3.0 / (1.0 + math.exp(( -60.0 - parameters[3]) * 0.050000000000000003))) * (1.1200000000000001 / (1.0 + math.exp((parameters[3] - 60.0) * 0.050000000000000003)))) # per_millisecond
+    {%- for eq in rhs_equations %}
+    cdef double {{eq}}
+    {%- endfor %}
+
+    # Pack state variable derivatives
+    {%- for state in states %}
+    (<Sundials.N_VectorContent_Serial>ydot.content).data[{{state[0]}}] = {{state[1]}}
```
jonc125: I think you want `2` not `1` :)
Another argument for `state.deriv_name` being better!

fccodegen/tests/test_model_generation.py
```diff
@@ -0,0 +1,30 @@
...
+    model = cg.load_model(model)
+
+    # Select output path (in temporary dir)
+    path = tmp_path / 'model.pyx'
+
+    # Create weblab model at path
+    cg.create_weblab_model(model, path)
+
+    # To see the result, use pytest -s
+    print(path.read_text())
```
jonc125: There should be an `assert` here to check the result against a (manually checked) reference version!

fccodegen/tests/test_printer.py
```diff
@@ -0,0 +1,197 @@
...
+    assert p.doprint(-x / x) == '-1'            # Negative one
+    assert p.doprint(5 * (x / x)) == '5'        # Sympy integer
+    assert p.doprint(5.5 * (x / x)) == '5.5'        # Sympy float
+    assert p.doprint(sp.Rational(5, 7)) == '5 / 7'  # Sympy rational
+
+    # Special numbers
+    assert p.doprint(sp.pi) == 'math.pi'
+    assert p.doprint(sp.E) == 'math.e'
+
+    # Symbols
```
jonc125: These things should really be split into multiple test methods. I don't go as far as saying a single `assert` per method, but a test method per category would be good.

fccodegen/tests/test_printer.py
```diff
@@ -0,0 +1,197 @@
...
+        (0, x > 0), (1, x != x), (2, True), (3, x > 3),
+        evaluate=False)
+    assert p.doprint(e) == '((0) if (x > 0) else (2))'
+
+    # Longer expressions
+    assert (
+        p.doprint((x + y) / (2 + z / sp.exp(x - y))) ==
+        '(x + y) / (2 + z * math.exp(y - x))')
+    assert p.doprint((y + sp.sin(x))**-1) == '1 / (y + math.sin(x))'
+
```
jonc125: Nice thorough lot of low-level tests!

fccodegen/tests/test_printer.py
```diff
@@ -0,0 +1,197 @@
...
+    # Unsupported sympy item
+    e = sp.Matrix()
+    with pytest.raises(RuntimeError):
```
jonc125: Can we have a more specific exception thrown please? `ValueError` might be a good candidate.

fccodegen/tests/test_templating.py
```diff
@@ -3,23 +3,46 @@
...
 
 
-def test_weblab_model(tmp_path):
+def test_unique_name_generation():
+    # Tests if unique variable names are generated correctly
+
+    # Load cellml model, get unique names
+    model = os.path.join(
+        cg.DIR_DATA, 'tests',
+        #'hodgkin_huxley_squid_axon_model_1952_modified.cellml'
```
jonc125: Remove line :)

fccodegen/_generator.py
```diff
@@ -1,26 +1,180 @@
...
+        if name in reverse:
+
+            # Get existing variable
+            other = reverse[name]
+
+            # Check it hasn't been renamed already
+            if symbols[other] == name:
+                oparts = other.name.split('$')
+                assert len(oparts) == 2
+                oname = uname(oparts[0] + '_' + oparts[1])
```
jonc125: This may be a personal preference thing, but I insert `__` between component & variable names - it makes it easier to distinguish when looking at generated code (especially given the propensity for underscores in component names) and reduces the likelihood of name conflicts like in your nasty test case.

fccodegen/_generator.py
```diff
@@ -1,26 +1,180 @@
...
+                symbols[other] = oname
+                reverse[oname] = other
+
+            # Get new name for v
+            name = uname(parts[0] + '_' + parts[1])
```
jonc125: Perhaps define `sep = '__'` and use that both here and in line 108 above?

fccodegen/_printer.py
```diff
@@ -0,0 +1,400 @@
...
+
+    def __init__(self, symbol_function=None, derivative_function=None):
+        super(WebLabPrinter, self).__init__(None)
+
+        # Prefix for functions
+        self._prefix = 'math.'
+
+        # Symbol and derivative handling (default)
+        if symbol_function is None:
+            self._symbol_function = lambda x: str(x)
```
jonc125: You should be able to use `str` directly rather than the `lambda`

fccodegen/_printer.py
```diff
@@ -0,0 +1,400 @@
...
+            self._symbol_function = symbol_function
+
+        if derivative_function is None:
+            self._derivative_function = lambda x: str(x)
+        else:
+            self._derivative_function = derivative_function
+
+    def _bracket(self, expr, parent_precedence):
+        """
+        Converts ``expr`` to string, and adds parentheses around the resut, if
```
jonc125: s/resut/result/

fccodegen/_printer.py
```diff
@@ -0,0 +1,400 @@
...
+        """
+        Called by :class:`Printer` as a last resort for unknown expressions.
+        """
+        raise RuntimeError(
+            'Unsupported expression type (' + str(type(expr)) + '): '
+            + str(expr))
+
+    def _print_Add(self, expr):
+        """ Handles addition & subtraction, with n terms. """
+        # This method is mostly copied from sympy.printing.Str
```
jonc125: Re 'mostly' - what's different?

fccodegen/_printer.py
```diff
@@ -0,0 +1,400 @@
...
+    def _print_Mul(self, expr):
+        """
+        Handles multiplication & division, with n terms.
+
+        Division is specified as a power: ``x / y --> x * y**-1``.
+        Subtraction is specified as ``x - y --> x + (-1 * y)``.
+        """
+        # This method is mostly copied from sympy.printing.Str
+
+        # Check overal sign of multiplication
```
jonc125: s/overal/overall/

fccodegen/_printer.py
```diff
@@ -0,0 +1,400 @@
...
+        """ Handles pi """
+        return self._prefix + 'pi'
+
+    def _print_Piecewise(self, expr):
+        """
+        Handles Piecewise functions.
+
+        Sympy's piecewise is defined as a list of tuples ``(expr, cond)`` and
+        evaluated by returning the first ``expr`` whose ``cond`` is true. If
+        non of the conditions hold a value error is raised.
```
jonc125: s/non/none/

fccodegen/_generator.py
MichaelClerx: It _is_ listed for cellmlmanip! The note is just to remind me that if we _don't_ move this code to cellmlmanip, I need to add networkx to fccodegen's dependencies

fccodegen/_generator.py
MichaelClerx: Nope! That only holds for a variable with an uncontested name. The first time we find that name "x" has been taken, we rename the old "x" so that its variable maps to "comp1_x", but we leave the "x": var mapping to indicate it's a contested name. The second time we find that name "x" has been taken (so a third contender for the name x), we find that, even though the variable is listed under "x", it calls itself "comp1_x" now, so we don't need to rename it a second time

fccodegen/_generator.py
MichaelClerx: Agreed!

fccodegen/templates/wl/weblab_model.pyx
MichaelClerx: Shall I use a `namedtuple`?

fccodegen/_generator.py
MichaelClerx: @tamuri would you like me to document these somewhere, or shall we just do a review of fccodegen together some time and fix it on the fly?

fccodegen/tests/test_printer.py
MichaelClerx: Thanks! I hope it's thorough, it's definitely a lot :-)

fccodegen/_generator.py
MichaelClerx: It can but that prints equality "==" rather than assignment "="!

fccodegen/__init__.py
```diff
@@ -14,10 +14,13 @@
...
     # https://docs.python.org/2/library/inspect.html#the-interpreter-stack
     del(frame)
+del(inspect)    # Don't expose imported modules
 
 # Template directory
 DIR_TEMPLATE = os.path.join(DIR_ROOT, 'templates')
-del(inspect)    # Don't expose as part of codegen
+
+# Directory for any other data
+DIR_DATA = os.path.join(DIR_ROOT, 'data')
```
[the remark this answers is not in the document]
MichaelClerx: Done!

fccodegen/templates/wl/weblab_model.pyx
MichaelClerx: Agreed!

fccodegen/tests/test_model_generation.py
MichaelClerx: Done!

fccodegen/tests/test_printer.py
MichaelClerx: Done!

fccodegen/_generator.py
jonc125: Ah, I see. That makes sense; I hadn't noticed that the old reverse mapping was kept too.

fccodegen/_generator.py
jonc125: Alas!

fccodegen/templates/wl/weblab_model.pyx
jonc125: You could do I guess, but I was just thinking of filling the `states` list using
```python
        states.append({'index': i, 'var_name': symbol_name(state), 'deriv_name': derivative_name(state)})
```
which I think should do the trick to enable dot-lookup name access.

fccodegen/tests/test_printer.py
jonc125: Nice use of fixtures!

fccodegen/templates/wl/weblab_model.pyx
MichaelClerx: Is that a jinja thing? Will give it a go!

fccodegen/_generator.py
MichaelClerx: Done

fccodegen/_printer.py
MichaelClerx: Done!

fccodegen/tests/test_templating.py
MichaelClerx: Done!

fccodegen/_generator.py
MichaelClerx: Done!

fccodegen/_printer.py
MichaelClerx: There's some order preserving thing that we don't have, and the variable names are slightly different, plus some if vs ternary stuff at the bottom. I've changed it to say it's based on sympy's one

MichaelClerx: Also I've added in comments to explain what the different bits of code do!

I always like it if the comments indicate what the code below is supposed to do (that way you can check if it does + it's more readable)

fccodegen/tests/test_printer.py
MichaelClerx: Done!

fccodegen/data/tests/weblab_model.pyx
```diff
@@ -3,31 +3,39 @@
...
+                      Sundials.N_Vector ydot,
+                      void* user_data):
     """
-    Main model method: evaluates the model's RHS.
+    Cython wrapper around a model RHS that uses numpy, for calling by CVODE.
+
+    See :meth:`fc.sundials.solver._EvaluateRhs()`.
     """
-    #TODO: What's this?
+    # Cast user data point from sundials back to numpy parameter vector.
```
jonc125: I'd use instead
```
# We passed the Python model object in as CVODE user data; get it back as an object
```

fccodegen/data/tests/weblab_model.pyx
```diff
@@ -72,36 +82,92 @@ cdef int _EvaluateRhs(Sundials.realtype var_environment__time, Sundials.N_Vector
...
+    # returned by GetOutputs
+    # From: fc.simulations.AbstractOdeModel
     cdef public object outputNames
 
+    # Mapping from names to saved model states.
+    # From: fc.simulations.AbstractOdeModel
     cdef public object savedStates
+
+    # An instance of fc.utility.environment.ModelWrapperEnvironment
+    # TODO: What's this?
```
jonc125: It maps names as used by the protocol to the corresponding variables in the model.

fccodegen/data/tests/weblab_model.pyx
```diff
@@ -72,36 +82,92 @@ cdef int _EvaluateRhs(Sundials.realtype var_environment__time, Sundials.N_Vector
...
+    # From: fc.simulations.AbstractOdeModel
     cdef public object env
+
+    # True if the solver needs to be reset due to a model change.
+    # TODO This is only ever set via the "environment". Not sure why.
+    # From: fc.simulations.AbstractOdeModel
     cdef public bint dirty
+
+    # TODO Not documented in AbstractModel
```
jonc125: Where to write protocol outputs, error logs, etc.

fccodegen/data/tests/weblab_model.pyx
```diff
@@ -72,36 +82,92 @@ cdef int _EvaluateRhs(Sundials.realtype var_environment__time, Sundials.N_Vector
...
+    # From: fc.simulations.AbstractOdeModel
     cdef public object savedStates
+
+    # An instance of fc.utility.environment.ModelWrapperEnvironment
+    # TODO: What's this?
+    # From: fc.simulations.AbstractOdeModel
     cdef public object env
+
+    # True if the solver needs to be reset due to a model change.
+    # TODO This is only ever set via the "environment". Not sure why.
```
jonc125: Because it's the `ModelWrapperEnvironment` that changes model variables from the protocol, and hence needs to flag to CVODE that a change has happened.

fccodegen/data/tests/weblab_model.pyx
```diff
@@ -72,36 +82,92 @@ cdef int _EvaluateRhs(Sundials.realtype var_environment__time, Sundials.N_Vector
...
+    # From: fc.simulations.AbstractModel
     cdef public char* outputPath
+
+    # Level of indentation to use for progress output.
+    # From: fc.simulations.AbstractModel
     cdef public object indentLevel
 
+    # Link to generated module.
+    # Set in: fc.utility.protocol.Protocol
+    # TODO: Nobody seems to ever access this variable. Is it just to prevent
```
jonc125: I think so. That code was written in 2013 :)

fccodegen/data/tests/weblab_model.pyx
```diff
@@ -72,36 +82,92 @@ cdef int _EvaluateRhs(Sundials.realtype var_environment__time, Sundials.N_Vector
...
+    # garbage collection?
     cdef public object _module
+
+    # TODO What's this?
```
jonc125: It's the environment for the simulation running this model, which is mainly useful when evaluating `set variable` type modifiers during the course of a simulation.

fccodegen/data/tests/weblab_model.pyx
```diff
@@ -193,3 +278,64 @@ cdef class GeneratedModel_Proto_tmpHzG5bQ(CvodeSolver):
...
+            self.cvode_mem, _EvaluateRhs, 0.0, self._state)
+        self.CheckFlag(flag, "CVodeInit")
+
+    def SetSolver(self, solver):
+        """
+        Specify the ODE solver to use for this model.
+
+        See :meth:`fc.simulations.AbstractOdeModel.SetSolver()`.
+        """
+        # TODO Should this be python 3 syntax? Or is this special Cython
```
jonc125: The `fc` code is Python 2 at present, and needs updating...

fccodegen/data/tests/weblab_model.pyx
```diff
@@ -193,3 +278,64 @@ cdef class GeneratedModel_Proto_tmpHzG5bQ(CvodeSolver):
...
+        # syntax?
+        # TODO Can we use logging here instead? Or an exception?
```
jonc125: Those would be better options, yes. I can't remember why this doesn't raise; probably there's some test cases that trigger this path.

MichaelClerx:
Hi @jonc125 . Would you like to review the updated code generation? I think 90% of it is there now?

fccodegen/__init__.py
```diff
@@ -9,19 +9,18 @@
 from ._config import (   # noqa
     DIR_DATA,
     DIR_ROOT,
-    DIR_TEMPLATE,
+    SUBDIR_TEMPLATE,
```
jonc125: 'Tis a very minor thing, but my personal preference would be to call these `DATA_DIR`, `ROOT_DIR` and `TEMPLATE_DIR`. I thought for a moment this was a template for subdirectories...

fccodegen/_generator.py
```diff
@@ -9,16 +9,45 @@
...
+    #TODO: Check absolute path is still in cg.DIR_TEMPLATE
+
+    env = _jinja_environment()
+    return env.get_template(path)
+
     #TODO: Check exists
     #TODO: Look at the jinja PackageLoader. What's the advantage?
-    with open(path, 'r') as f:
-        return jinja2.Template(f.read())
+    #with open(path, 'r') as f:
```
jonc125: Get rid of these rather than commenting (and the TODO above :) )

fccodegen/_generator.py
```diff
@@ -38,21 +67,37 @@ def load_model(path):
...
 
     # Get sorted symbols
-    #TODO: networkx is not a dependency
+    #TODO: networkx is not a dependency of fccodegen - only of cellmlmanip. If
+    #      this does _not_ get moved to cellmlmanip, it should become a dep.
+    #      here too.
     import networkx as nx
-    symbols = nx.lexicographical_topological_sort(graph, key=lambda x: str(x))
+    sorted_symbols = nx.lexicographical_topological_sort(
+        graph, key=lambda x: str(x))
```
jonc125: You should be able to do `key=str`

fccodegen/_generator.py
```diff
@@ -38,21 +67,37 @@ def load_model(path):
...
+
+    # Create set of symbols for which we require equations
```
jonc125: It'll be more efficient (particularly for outputs) if you can do a topological sort just on the sub-graph that's actually required. But save implementing this for `cellmlmanip`.

fccodegen/_generator.py
```diff
@@ -158,39 +273,86 @@ def derivative_name(deriv):
...
             'index': i,
             'var_name': symbol_name(state),
             'deriv_name': derivative_name(state),
+            'initial_value': initial_value,
+        })
+
+    # Create parameter information dicts
+    parameter_info = []
+    for i, parameter in enumerate(parameters):
+        symbol = get_symbol_by_cmeta_id(graph, parameter)
```
jonc125: This will need to be an RDF lookup eventually, but I think you're aware of that :)

fccodegen/_generator.py
```diff
@@ -158,39 +273,86 @@ def derivative_name(deriv):
...
+        parameter_info.append({
+            'index': i,
+            'cmeta_id': parameter,
+            'var_name': symbol_name(symbol),
+            'initial_value': get_value(graph, symbol),
+        })
+
+    # Create map of parameter symbols to their indices
+    parameter_symbols = {}
+    for i, parameter in enumerate(parameters):
```
jonc125: Is there a reason not to fold this into the loop above?

fccodegen/_generator.py
```diff
@@ -158,39 +273,86 @@ def derivative_name(deriv):
...
+        symbol = get_symbol_by_cmeta_id(graph, parameter)
+        parameter_symbols[symbol] = i
+
+    # Create output information dicts
+    # Each output is associated either with a symbol or a parameter.
```
jonc125: Presumably the symbol case handles when the free variable is output?

fccodegen/_generator.py
```diff
@@ -158,39 +273,86 @@ def derivative_name(deriv):
...
+    # Create RHS equation information dicts
     rhs_equations = []
-    for eq in eqs:
-        lhs, rhs = eq.lhs, eq.rhs
-        rhs_equations.append(p.doprint(lhs) + ' = ' + p.doprint(rhs))
+    for eq in get_equations_for(graph, get_derivative_symbols(graph)):
+        rhs_equations.append({
+            'lhs': printer.doprint(eq.lhs),
+            'rhs': printer.doprint(eq.rhs),
+            'parameter_index': parameter_symbols.get(eq.lhs, None),
```
jonc125: You shouldn't ever have (once the manipulation is done) a parameter occurring on the LHS of an equation. They can only be set from the protocol, not from a model equation.

fccodegen/_generator.py
```diff
@@ -158,39 +273,86 @@ def derivative_name(deriv):
...
         f.write(template.render({
-            'date': time.strftime('%Y-%m-%d %H:%M:%S'),
-            'generator': cg.version(formatted=True),
-            'name': model.name,
+            'class_name': class_name,
+            'free_variable': free_name,
+            'generation_date': time.strftime('%Y-%m-%d %H:%M:%S'),
+            'generator_name': cg.version(formatted=True),
+            'model_name': model.name,
+            'n_parameters': len(parameter_info),
```
jonc125: You can get this from the `parameter_info` variable within the template using the `length` [filter](http://jinja.pocoo.org/docs/2.10/templates/#filters) - no need to pass as a separate variable.

fccodegen/data/tests/weblab_model.pyx
```diff
@@ -158,42 +156,43 @@ cdef class GeneratedModel_Proto_tmpHzG5bQ(CvodeSolver): # TODO Generate name
         self.freeVariableName = "time"
         self.freeVariable = 0.0
 
-        # TODO Generate this
-        self.state = np.zeros(2)
+        # State values
+        self.state = np.zeros(4)
+
+        # Mapping from cmeta ids to state indices
+        # TODO Is this correct? Not all states need to have a cmeta id!
```
jonc125: The docstring says "stateVarMap: a mapping from variable name to index within the state variable vector, for use in our ModelWrapperEnvironment". Variable name here is the name as known by the protocol, i.e. oxmeta name rather than cmeta:id. A state variable which isn't annotated therefore doesn't appear in this map and can't be altered (or read individually) from the protocol.

fccodegen/tests/test_weblab_templating.py
```diff
@@ -36,7 +52,7 @@ def test_generate_weblab_model(tmp_path):
     generated = path.read_text()
 
     # Store locally to update test output file
-    if True:
+    if False:
```
jonc125: I now tend to make lines like this read an environment variable, so I can just do e.g. `WEBLAB_REGENERATE_REF=1 pytest` to regenerate all reference data.

fccodegen/_generator.py
tamuri: Might be my fault in cellmlmanip...?

jonc125: You have the right definition of a parameter for FC. However cellmlmanip will also be used to do the model manipulations specified by a protocol, so there will be some overlap. Once everything is connected up, the FC code would use cellmlmanip to modify the model equations, which will (inter alia) get rid of all equations with (FC) parameters on the LHS (they'll then only appear on the RHS of other equations), and then use fccodegen to generate the code to simulate the model.

fccodegen/_generator.py
MichaelClerx: Yeah it's a silly implementation! But I just wanted something that'd work with the 5 API calls I knew about :-)

fccodegen/__init__.py
MichaelClerx: Done!

fccodegen/_generator.py
MichaelClerx: ok!

fccodegen/_generator.py
MichaelClerx: D'oh!

fccodegen/_generator.py
MichaelClerx: Have made a note of it in the code too now!

fccodegen/_generator.py
MichaelClerx: Readability! (There should be no appreciable performance hit)

fccodegen/_generator.py
MichaelClerx: Yup! I test with that too

fccodegen/_generator.py
MichaelClerx: I'm still not sure how/where the manipulations will be applied! The protocol language seems to modify cellml, which would imply changes pre-parsing. Cellmlmanip removes cellml as soon as it can rather than actually manipulating it, right?

fccodegen/_generator.py
MichaelClerx: I think this is cleaner, no?

fccodegen/data/tests/weblab_model.pyx
MichaelClerx: Thanks!

fccodegen/_generator.py
jonc125: The protocol language modifies equations, not CellML. So those changes happen post-parsing. (This is a key difference from SED-ML, which modifies the raw XML.)

fccodegen/tests/test_weblab_templating.py
MichaelClerx: Done!

fccodegen/_generator.py
jonc125: I disagree :) It gives more scope for things to get out of sync (small scope, admittedly, but why pass a new variable you don't need?)

fccodegen/_generator.py
MichaelClerx: Thanks! Have made a note

PR #5, opened 2019-01-25
Fixes to web lab template.

MichaelClerx:


fccodegen/data/tests/weblab_model.pyx
```diff
@@ -145,6 +145,9 @@ cdef class TestModel(CvodeSolver):
     # prevent garbage collection.
     cdef public object _module
 
+    # NOT SURE
```
jonc125: From memory this was a bit of a hack, as a way of referring to the environment (i.e. variable mapping) of the simulation currently running the model. Searching for `simEnv` in the fc code should yield some more context.

MichaelClerx: Oh yeah, there's some to-dos about removing it somehow :D

PR #6, opened 2019-02-18
Replace graph manipulating code with new `cellmlmanip` API methods

MichaelClerx:


MichaelClerx:
Only bit of internal `graph` usage left at this point is the method to define unique variable names. We already decided this was a code generation issue rather than a cellmlmanip feature, so probably OK for it to be a bit more involved?

I think this is ready for review!

fccodegen/_generator.py
```diff
@@ -341,9 +205,9 @@ def derivative_name(deriv):
 
     # Create output equation information dicts
     output_equations = []
-    output_symbols = [get_symbol_by_cmeta_id(graph, x) for x in outputs
+    output_symbols = [model.get_symbol_by_cmeta_id(x) for x in outputs
                       if x != 'state_variable']
```
jonc125: I take it `state_variable` is handled specially elsewhere?

MichaelClerx: Yeah that's handled here: https://github.com/ModellingWebLab/fccodegen/blob/cellmlmanip-updates/fccodegen/_generator.py#L174-L180

PR #8, opened 2019-03-01
Annotation lookups

MichaelClerx:
@jonc125 were you thinking of something like this? Or would we still use cmeta_ids in the generated code?

It also seems like it might be a good idea to introduce some kind of "annotation" object here? (Basically a tuple `{namespace_uri}local_name`). Or perhaps use rdflib nodes?


weblab_cg/data/tests/static_model_graphstate.pyx
```diff
@@ -181,8 +181,8 @@ cdef class TestModel(CvodeSolver):
 
         # Oxmeta names of output variables
         self.outputNames = []
-        self.outputNames.append('time')
-        self.outputNames.append('membrane_voltage')
+        self.outputNames.append('https://chaste.comlab.ox.ac.uk/cellml/ns/oxford-metadata#time')
```
jonc125: WL1 uses just the local names here, without the base URI part. What we should do eventually is update the `ModelWrapperEnvironment` so we can use a separate instance for each namespace defined by the protocol, and then we can use longer names here and let each environment wrap its respective subset. But until that happens, users just have to make sure not to use the same local name in different namespaces.

jonc125: Note that these should be local names *not* cmeta ids, even though they're sometimes the same!

MichaelClerx:
Have changed the `outputNames` back to local names now, and fixed the bad commit message

MichaelClerx:
Could you have another look please @jonc125 ?

weblab_cg/data/tests/weblab_model.pyx
```diff
@@ -173,17 +173,17 @@ cdef class TestModel(CvodeSolver):
         self.initialState[2] = 0.6
         self.initialState[3] = 0.325
 
-        # Mapping of parameter oxmeta names to parameter array indices
+        # Mapping of parameter qualified names to parameter array indices
         self.parameterMap = {}
-        self.parameterMap['membrane_fast_sodium_current_conductance'] = 0
-        self.parameterMap['membrane_potassium_current_conductance'] = 1
+        self.parameterMap['https://chaste.comlab.ox.ac.uk/cellml/ns/oxford-metadata#membrane_fast_sodium_current_conductance'] = 0
```
jonc125: Should be just local names here I think?

weblab_cg/templates/wl/weblab_model.pyx
```diff
@@ -152,10 +152,10 @@ cdef class {{ class_name }}(CvodeSolver):
         self.initialState[{{ state.index }}] = {{ state.initial_value }}
         {%- endfor %}
 
-        # Mapping of parameter oxmeta names to parameter array indices
+        # Mapping of parameter qualified names to parameter array indices
         self.parameterMap = {}
         {%- for parameter in parameters %}
-        self.parameterMap['{{ parameter.cmeta_id }}'] = {{ parameter.index }}
+        self.parameterMap['{{ parameter.annotation[0] + parameter.annotation[1] }}'] = {{ parameter.index }}
```
jonc125: And here...

weblab_cg/data/tests/static_model_graphstate.pyx
MichaelClerx: Thanks, have put these comments in the source code for now

weblab_cg/data/tests/weblab_model.pyx
MichaelClerx: Done!

weblab_cg/templates/wl/weblab_model.pyx
MichaelClerx: Done!

PR #19, opened 2019-08-15
Update README.md

MichaelClerx:


MichaelClerx:
@tamuri Some changes in cellmlmanip that break weblab_cg seem to have crept in, perhaps as part of the sympy update? Any ideas?

tamuri:
Sympy changed the interface to some basic expressions. Check anything that works on sympy.Derivative - the argument types have changed so now the second argument is a sympy.Tuple type like (sympy.Symbol, sympy.Integer)

MichaelClerx:
> Was this just fixing the model/module type? I can't see anything else.

Yeah! Bit of a downside to doing everything through PRs :p

MichaelClerx:
> Sympy changed the interface to some basic expressions. Check anything that works on sympy.Derivative - the argument types have changed so now the second argument is a sympy.Tuple type like (sympy.Symbol, sympy.Integer)

Thanks! So is 1.4 the minimum version now?
I see some derivative changes here: https://github.com/sympy/sympy/wiki/Release-Notes-for-1.4

tamuri:
> So is 1.4 the minimum version now?

I only tested on 1.4

PR #28, opened 2019-09-04
use flake8 settings from cellmlmanip

MauriceHendrix:
imported just flake8 settings from cellmlmanip as something in the other settings seems to break things.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,33 +1,7 @@
 [flake8]
-exclude=
-    .git,
-    dev-requirements,
-    venv,
-ignore=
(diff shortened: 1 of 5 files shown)


.flake8
```diff
@@ -1,33 +1,7 @@
 [flake8]
-exclude=
-    .git,
-    dev-requirements,
-    venv,
```
MichaelClerx: Should keep these includes in there (a) to speed things up, and (b) so that we don't style-check things installed into virtual environments

MauriceHendrix: done

PR #29, opened 2019-09-16
Update generated FC code to use new syntax

MichaelClerx:
See https://github.com/ModellingWebLab/weblab-fc/pull/35


MichaelClerx:
Note this only changes a template + a static output file for testing. No changes to functional parts of cg

PR #35, opened 2019-11-05
Chaste cg

MauriceHendrix:
Chaste codegenenratio first steps.
I'm aware it needs more/better doc strings (working on it). Also some things could perhaps be done better most specifically:
_format_stimulus_currents
_format_extended_equations_for_ionic_vars(self):
_format_derivative_equations


.gitignore
```diff
@@ -104,5 +104,8 @@ ENV/
 /.idea
 *.iml
 
+# Pytest cache
+.pytest_cache
```
MichaelClerx: This doesn't seem to be working? There are cache files just below!

results.txt
```diff
@@ -0,0 +1,99 @@
+============================= test session starts =============================
```
MichaelClerx: Pretty sure this shouldn't be here

weblab_cg/__init__.py
```diff
@@ -29,7 +29,21 @@
...
+    NormalChasteModel,
+    OptChasteModel,
+    Analytic_jChasteModel,
+    Numerical_jChasteModel,
+    BEOptChasteModel,
+)
+
 from ._printer import (     # noqa
     WebLabPrinter,
+    _math_functions,
```
MichaelClerx: Shouldn't usually import hidden things...

results.txt
MauriceHendrix: yes sorry

.gitignore
MauriceHendrix: Seems to be from before I have deleted it.

weblab_cg/__init__.py
MauriceHendrix: Fair point I made it a class level variable in the printer so I have access to it

weblab_cg/__init__.py
```diff
@@ -29,7 +29,20 @@
     load_template,
 )
 
+from ._generator_chaste_cg import (   # noqa
+    NormalChasteModel,
+    OptChasteModel,
```
MichaelClerx: Can we use more descriptive and more consistent names? What does `Opt` mean? And BEOpt? And why mix underscores with CamelCase?

weblab_cg/_chaste_printer.py
```diff
@@ -0,0 +1,91 @@
...
+    """
+
+    def __init__(self, symbol_function=None, derivative_function=None):
+        super().__init__(symbol_function, derivative_function)
+
+        # Do not use the math. prefix for functions, as chaste c++ files include an include for this
+        self._prefix = ''
+
+        # Make sure we can output a call to GetIntracellularAreaStimulus
+        WebLabPrinter._math_functions.add('GetIntracellularAreaStimulus')
```
MichaelClerx: Bit confused about this. `_math_functions` is a list of functions that exist in both sympy and python's `math` module, and so can be passed through directly. (1) these names are not necessarily the same in C++ and (2) there's certainly no GetIntracellularAreaStimulus in sympy

weblab_cg/_chaste_printer.py
```diff
@@ -0,0 +1,91 @@
+from ._printer import WebLabPrinter
+import sympy
+
+from sympy.printing.precedence import precedence
```
MichaelClerx: Is sympy operator precedence the same as C++ precedence?

weblab_cg/_chaste_printer.py
```diff
@@ -0,0 +1,91 @@
...
+
+    def _print_Or(self, expr):
+        """ Handles logical Or. """
+        my_prec = precedence(expr)
+        return ' || '.join([self._bracket(x, my_prec) for x in expr.args])
+
+    def _print_Pow(self, expr):
+        """ Handles Pow(), which includes all division
+        only ordinary power is different, the rest is handed back up to the parent class (WebLabPrinter) """
+        # Square root or ((only if communicative) 1/sqrt or ordinary devision )
```
MichaelClerx: I don't know what this comment means. Chances are you won't remember in a year either :D

weblab_cg/_chaste_printer.py
```diff
@@ -0,0 +1,91 @@
...
+        if expr.exp is sympy.S.Half or (expr.is_commutative and
+                                        (-expr.exp is sympy.S.Half or -expr.exp is sympy.S.One)):
+            return super()._print_Pow(expr)
+        else:
+            # Ordinary power
+            p = precedence(expr)
```
MichaelClerx: You definitely don't need brackets here: `pow(x, y)` vs `pow((x), (y))`

weblab_cg/_chaste_printer.py
```diff
@@ -0,0 +1,91 @@
...
+            return 'pow(' + self._bracket(expr.base, p) + ', ' + self._bracket(expr.exp, p) + ')'
```
MichaelClerx: Maybe we should update the WeblabPrinter to have some method `_print_ordinary_pow()` or something that you could override here, instead of having to implement the long condition above (`if expr.exp` etc) in every subclass?
Seems to me that long condition is the hardest part of this method, so don't want to repeat it where it isn't needed

weblab_cg/_chaste_printer.py
```diff
@@ -0,0 +1,91 @@
...
+        other = 'float(\'nan\')'
+
+        parts = ['(']
+        brackets = 1
+        for e, c in expr.args:
+            # Check if boolean True (if found, stop evaluating further)
+            if isinstance(c, BooleanTrue):
+                other = self._print(e)
+                break
+            parts.append('(')
```
MichaelClerx: Maybe we can have a `_print_ternary(cond, iftrue, iffalse)` function in WebLabPrinter that we can override here?

weblab_cg/__init__.py
```diff
@@ -29,7 +29,20 @@
...
+    Analytic_jChasteModel,
+    Numerical_jChasteModel,
+    BEOptChasteModel,
+)
+
 from ._printer import (     # noqa
```
MichaelClerx: Should this now we from `._weblab_printer` or something? Should we have a module per output format?

CONTRIBUTING.md
```diff
@@ -83,3 +83,25 @@ These docstrings can be fairly simple, but can also make use of [reStructuredTex
 ## Infrastructure & configuration files
 
 TODO
+
+
+### Visual Studio Code Development Environment
```
MichaelClerx: These are good notes, but very specific for one particular IDE. Are we going to add notes for every IDE?

weblab_cg/tests/test_chaste_printer.py
```diff
@@ -0,0 +1,55 @@
+#
+# Tests conversion of sympy expressions to weblab cython code.
```
MichaelClerx: No it doesn't :D

weblab_cg/templates/chaste/normal_model.hpp
```diff
@@ -0,0 +1,83 @@
+#ifndef {% filter upper %}{{class_name}}FROMCELLML_HPP_{% endfilter %}
+#define {% filter upper %}{{class_name}}FROMCELLML_HPP_{% endfilter %}
+
+//! @file
+//! 
+//! This source file was generated from CellML.
+//! 
+//! Model: {{model_name_from_file}}
+//! 
+//! Processed by webalab_cg - CellML Tools in Python
```
MichaelClerx: There's lots of CellML tools in python!

Maybe we could put a link to the repo or something?

weblab_cg/templates/chaste/normal_model.hpp
```diff
@@ -0,0 +1,83 @@
...
+//!     (translators: , webalab_cg: , options: normal)
+//! on {{generation_date}}
+//! 
+//! <autogenerated>
```
MichaelClerx: What's this? I'm feeling uneasy about the lack of a closing tag

weblab_cg/templates/chaste/normal_model.cpp
```diff
@@ -0,0 +1,148 @@
...
+    this->mSystemName = "{{free_variable.system_name}}";
+    this->mFreeVariableName = "{{free_variable.name}}";
+    this->mFreeVariableUnits = "{{free_variable.units}}";
+    
+    {% for ode_info in ode_system_information %}// rY[{{loop.index0}}]:
+    this->mVariableNames.push_back("{{ode_info.name}}");
+    this->mVariableUnits.push_back("{{ode_info.units}}");
+    this->mInitialConditions.push_back({{ode_info.initial_value}});
+	
+	{% endfor %}{% for attr in named_attributes %}// rY[{{loop.index0}}]:
```
MichaelClerx: Leftover debug code?

weblab_cg/templates/chaste/normal_model.cpp
```diff
@@ -0,0 +1,148 @@
...
+    }
+    
+template<>
+void OdeSystemInformation<{{class_name}}FromCellML>::Initialise(void)
+{
+    this->mSystemName = "{{free_variable.system_name}}";
+    this->mFreeVariableName = "{{free_variable.name}}";
+    this->mFreeVariableUnits = "{{free_variable.units}}";
+    
+    {% for ode_info in ode_system_information %}// rY[{{loop.index0}}]:
```
MichaelClerx: Leftover debug code?

weblab_cg/templates/chaste/normal_model.cpp
```diff
@@ -0,0 +1,148 @@
...
+#include "RegularStimulus.hpp"
+#include "HeartConfig.hpp"
+#include "IsNan.hpp"
+#include "MathsCustomFunctions.hpp"
+	{%- if default_stimulus_equations.membrane_stimulus_current_duration is defined %}
+
+    boost::shared_ptr<RegularStimulus> {{class_name}}FromCellML::UseCellMLDefaultStimulus()
+    {
+        // Use the default stimulus specified by CellML metadata
+		{% for key, value in default_stimulus_equations.items() %}
```
MichaelClerx: Mixing tabs and spaces: Maybe check all files for this! (And see if you can get your IDE to warn you when this happens?)

weblab_cg/data/tests/chaste_reference_models/Normal/hodgkin_huxley_squid_axon_model_1952_modified.hpp
```diff
@@ -0,0 +1,77 @@
+#ifndef DYNAMICHODGKIN_HUXLEY_SQUID_AXON_MODEL_1952_MODIFIEDFROMCELLML_HPP_
```
MichaelClerx: This file is already in the `data` directory. We don't need to have it twice

weblab_cg/data/tests/chaste_reference_models/cellml/aslanidi_model_2009.cellml
```diff
@@ -0,0 +1,4092 @@
+<?xml version="1.0" encoding="utf-8"?>
```
MichaelClerx: Do we really need four separate files? If we do it properly, unit tests should have caught all functionality in nice little modular blocks, and big tests like this won't add any value. At the very least we could just select 1 big cellml file and edit it until it contains all the features we want? Or if there are very different ways of using a cellml file, then an example of each, but this just looks like 4 very similar testcases to me

weblab_cg/tests/test_chaste_cg.py
```diff
@@ -0,0 +1,528 @@
+#
+# Tests the basics of weblab_cg
```
MichaelClerx: Nope

weblab_cg/tests/test_chaste_cg.py
```diff
@@ -0,0 +1,528 @@
...
+logging.getLogger().setLevel(logging.DEBUG)
+# TODO: Better docstrings
+
+
+class TestChasteCG(object):
+    _COMMENTS_REGEX = re.compile(r'(//.*\n)')
+    _NUM_REGEX = re.compile(r'(?:0|[1-9]\d*)(?:\.\d*)?(?:[eE][+\-]?\d+)?')
+    _FLOAT_PRECISION = 11
+
+    def _make_dirs(self, output_path):
```
MichaelClerx: I don't think we need private methods in the tests, do we? Can these just be public?

weblab_cg/tests/test_chaste_cg.py
```diff
@@ -0,0 +1,528 @@
...
+import re
+import weblab_cg as cg
+import pytest
+import sympy
+from sympy import SympifyError
+import pyparsing
+
+# Show more logging output
+logging.getLogger().setLevel(logging.DEBUG)
+# TODO: Better docstrings
```
MichaelClerx: Yes please!
These look like very complicated tests, much more complex than I'm used to seeing in docstrings (are there no simpler tests of small chunks we could be doing?)
Would like a lot of comments to document what these tests are supposed to do, so that we read the comments, then read the code and check if it does what the comments tell us it should be doing (we can only tell what the code _does_ by reading it, not what it should do / what the intention was)

weblab_cg/data/tests/chaste_reference_models/Normal/hodgkin_huxley_squid_axon_model_1952_modified.hpp
MichaelClerx: (Sorry, I meant the cellml file)

weblab_cg/data/tests/chaste_reference_models/Normal/hodgkin_huxley_squid_axon_model_1952_modified.cpp
```diff
@@ -0,0 +1,170 @@
...
+		// Units: dimensionless; Initial value: 0.05
+		double var_chaste_interface__sodium_channel_h_gate__h = rY[2];
+		// Units: dimensionless; Initial value: 0.6
+		double var_chaste_interface__potassium_channel_n_gate__n = rY[3];
+		// Units: dimensionless; Initial value: 0.325
+        
+        // Mathematics
+		double d_dt_chaste_interface__membrane__V;
+		const double var_potassium_channel_n_gate__alpha_n = -0.01 * (65.0 + var_chaste_interface__membrane__V) / (-1.0 + 0.0015034391929775724 * exp(-0.1 * var_chaste_interface__membrane__V)); // per_millisecond
+		const double var_potassium_channel_n_gate__beta_n = 0.31919868225786585 * exp(0.0125 * var_chaste_interface__membrane__V); // per_millisecond
```
MichaelClerx: Is it numbers like these that were causing all the trouble with the testing? Would it be possible to just manually update these reference files to avoid having complex testing code?

weblab_cg/data/tests/chaste_reference_models/Normal/bondarenko_model_2004_apex.cpp
```diff
@@ -0,0 +1,1056 @@
+//! @file
```
MichaelClerx: Why is this directly capitalised all of a sudden? Just asking :D

weblab_cg/_printer.py
```diff
@@ -163,6 +161,13 @@ def _print_And(self, expr):
         my_prec = precedence(expr)
         return ' and '.join([self._bracket(x, my_prec) for x in expr.args])
 
+    def _print_bool(self, expr):
+        """ Handles False """
```
MichaelClerx: It doesn't just handle `False`

weblab_cg/_generator_chaste_cg.py
```diff
@@ -0,0 +1,627 @@
...
+                                                                    (1 / equations[i]['factor'] * equations[i]['lhs'])})
+            if float(equations[i]['factor']) != 1.0:
+                equations[i]['rhs'] = sp.simplify(equations[i]['factor'] * equations[i]['rhs'])
+            equations[i]['rhs'] = \
+                self._printer.doprint(equations[i]['rhs']) + equations[i]['rhs_divider']
+            equations[i]['lhs'] = self._printer.doprint(equations[i]['lhs'])
+
+        return equations
+
+    def _format_system_info(self):
```
MichaelClerx: - Maybe `ode_system` instead of `system`? I was expecting it to show python versions, operating system etc.
- The method doesn't format info (`format_info(self, info)`) but generates and returns info, in some complex type. Would update the name and docstring to describe what this thing returns.
- The code below isn't very easy to read so please add some comments explaining what it's supposed to do / what the expected output is

MichaelClerx: (as in, imagine something happens upstream so that this method stops working, and I want to update it, then I need to know what the expected output is)

weblab_cg/_generator_chaste_cg.py
```diff
@@ -0,0 +1,627 @@
+#
```
MichaelClerx: This looks pretty serious. Nice!

weblab_cg/_generator_chaste_cg.py
```diff
@@ -0,0 +1,627 @@
...
+        self._add_printers()
+        self._formatted_state_vars = self._format_state_variables()
+        self._formatted_default_stimulus = self._format_stimulus_currents()
+        self._formatted_equations_for_ionic_vars = self._format_equations_for_ionic_vars()
+        self._formatted_extended_equations_for_ionic_vars = self._format_extended_equations_for_ionic_vars()
+        self._formatted_y_derivatives = self._format_y_derivatives()
+        self._formatted_derivative_eqs = self._format_derivative_equations()
+        self._free_variable, self._ode_system_information, self._named_attributes = self._format_system_info()
+
+    def generate_chaste_code(self):
```
MichaelClerx: If I can't call it, then why is this here?

weblab_cg/_generator_chaste_cg.py
```diff
@@ -0,0 +1,627 @@
...
+                current_units = self._model.units.summarise_units(membrane_capacitance)
+                desired_units = getattr(self._model.units.ureg, self._STIMULUS_UNITS['membrane_capacitance'])
+                factor = self._model.units.get_conversion_factor(desired_units, from_unit=current_units)
+                initial_value = self._model.get_initial_value(membrane_capacitance)
+                equation = sp.Eq(membrane_capacitance, initial_value)
+                return equation, factor
+            except KeyError:
+                return None, 1.0
+
+    def _get_stimulus_currents(self):
```
MichaelClerx: Not sure I get this. Is there every more than 1 stimulus current?

weblab_cg/_generator_chaste_cg.py
```diff
@@ -0,0 +1,627 @@
...
+            return True
+        except KeyError:
+            return False
+
+    def _get_state_variables(self):
+        """ Sort the state variables, in similar order to pycml to prevent breaking existing code"""
+        return sorted(self._model.get_state_symbols(),
+                      key=lambda state_var: self._state_var_key_order(state_var))
+
+    def _get_membrane_stimulus_current(self):
```
MichaelClerx: These methods, `_get_physical_variable`, are all very useful and should probably go in `cellmlmanip` (maybe in a separate module like [I'm planning](https://github.com/MichaelClerx/myokit/issues/351) for Myokit?). Maybe without the default bit though, so just a get_x method that returns x or raise a KeyError or returns None or something

weblab_cg/_generator_chaste_cg.py
MichaelClerx: Should I change my generator to `weblab_generator` then? If they're not sharing any code?

weblab_cg/_chaste_printer.py
MauriceHendrix: Indeed that's why I added them! Maybe math_functions is the wrong name and we should change it to just functions. If I don't add these in as recognized functions the printer fails when I try to output them (telling me i'm writing out an unknown function).

weblab_cg/_chaste_printer.py
MauriceHendrix: As far as I can tell for maths expressions they should be no?

weblab_cg/_chaste_printer.py
MauriceHendrix: I've updated the docstring (will be in next commit)
        """ Handles Pow(), which includes all division
        
        for C++ printing we need to writea x**y as pow(x, y) with lowercase p
        only ordinary power is different from the weblab printer
        the rest (suqare root and division is handed back up to the parent class (WebLabPrinter) """

weblab_cg/_chaste_printer.py
MauriceHendrix: true

weblab_cg/templates/chaste/normal_model.cpp
MauriceHendrix: nope propper chaste codebeing generated (although the // rY[{{loop.index0}}]: is a mistake)

weblab_cg/_printer.py
MauriceHendrix: No maybe more useful to add that it's for python bools (done)

weblab_cg/_generator_chaste_cg.py
MauriceHendrix: To make sure you get a useful and descriptive error message  if you accidentally call it on the wrong type of model. Also I like to be able to see from a base class if there is a method all subclasses will have

weblab_cg/_chaste_printer.py
MichaelClerx: No idea! There could be small differences :-)
Maybe just check and add a comment above saying they're the same?

weblab_cg/templates/chaste/normal_model.cpp
MichaelClerx: That's the bit I meant :-D

weblab_cg/_chaste_printer.py
MauriceHendrix: I would if I could find the sympy precedence  but for the life of me I can't find it.

weblab_cg/templates/chaste/normal_model.cpp
MauriceHendrix: entierly legit chaste code (it's the template remember?)

weblab_cg/_chaste_printer.py
MichaelClerx: Haha, I suppose it's the same as for Python, but guessing isn't checking :p

weblab_cg/templates/chaste/normal_model.cpp
MichaelClerx: Oooooh I see it now!

weblab_cg/_chaste_printer.py
MauriceHendrix: Hhm it doesn't work without the brackets. Actually if you look at your own printer it also has:

`return self._bracket(expr.base, p) + '**' + self._bracket(expr.exp, p)`

weblab_cg/_chaste_printer.py
MichaelClerx: Hmmmm, but there's no sympy function that maps to GetIntracellularAreaStimulus, is there?

MauriceHendrix: As far as I understand you're just defining functions that are legal in the output. They don't have to by in sympy

MichaelClerx: It's used exclusively by `_print_Function(self, expr):`, which is called when a sympy function needs to be printed

weblab_cg/_chaste_printer.py
MauriceHendrix: good idea!

weblab_cg/_chaste_printer.py
MauriceHendrix: sure but more like `_print_ternary(self, cond, expr)` as it seems to come in cond, expr pairs from piecewise

CONTRIBUTING.md
MauriceHendrix: To be honest I mostly added them in case I end up having to get back into it after having done something else for a couple of months. I have updated the comment to indicate that text editors are fine and there are many IDEs

weblab_cg/templates/chaste/normal_model.hpp
MauriceHendrix: good idea!

weblab_cg/templates/chaste/normal_model.hpp
MauriceHendrix: Dood point. That's what chaste generated code looked like (no closing tag. Easy enough to add one though.

weblab_cg/data/tests/chaste_reference_models/Normal/hodgkin_huxley_squid_axon_model_1952_modified.hpp
MauriceHendrix: oeps, gone next update!

weblab_cg/templates/chaste/normal_model.hpp
mirams: I'd leave things like that in comments alone, it probably excludes the file from doxygen coverage testing or something in Chaste tests.

mirams: In general can we focus on comments that need addressing for stuff to work 😉 there is a cost to everything people get emails about and need to read and think about!

CONTRIBUTING.md
MichaelClerx: Haha, I was just worried it would get too long / out-of-scope if everyone started adding their own editor setups :D

weblab_cg/_generator_chaste_cg.py
MichaelClerx: Woops, sorry didn't notice this was a base class

MauriceHendrix: lol no worries

weblab_cg/templates/chaste/normal_model.hpp
MauriceHendrix: sure

weblab_cg/data/tests/chaste_reference_models/cellml/aslanidi_model_2009.cellml
MauriceHendrix: Eventually maybe not but for now yes. I've split the test into a develop test with the complicated symbolic test to help me develop and a straight textual comparison of models that I've already tested in chaste.

weblab_cg/_generator_chaste_cg.py
MauriceHendrix: could do, up to you really, but we're sharing quite a lot still (hence the sub classing)

weblab_cg/data/tests/chaste_reference_models/Normal/bondarenko_model_2004_apex.cpp
MauriceHendrix: I don't know what you mean. This version was pycml reference output, so not my file.

weblab_cg/data/tests/chaste_reference_models/Normal/hodgkin_huxley_squid_axon_model_1952_modified.cpp
MauriceHendrix: Yes and no (as a few occasions caused by simplify are also in the generated code) but I've found a way round it it's "only" for the develop test anyway.

weblab_cg/tests/test_chaste_cg.py
MauriceHendrix: Yes at some point, for now I've split between a simple direct comparison test and my development tool test

weblab_cg/_generator_chaste_cg.py
MauriceHendrix: hhm currents is maybe not the word. I'll just call it _get_stimulus
result looks like this:

`"{'membrane_stimulus_current_amplitude': Eq(_membrane$stim_amplitude, -20.0), 'membrane_stimulus_current_duration': Eq(_membrane$stim_duration, 0.0002), _membrane$stim_end: Eq(_membrane$stim_end, 100.0), 'membrane_stimulus_current_period': Eq(_membrane$stim_period, 0.5), 'membrane_stimulus_current_offset': Eq(_membrane$stim_start, 0.01)}"`

weblab_cg/tests/test_chaste_cg.py
MauriceHendrix: They can but I don't really see the benefit.

weblab_cg/tests/test_chaste_cg.py
MauriceHendrix: updated

weblab_cg/tests/test_chaste_printer.py
MauriceHendrix: ops, fixed

weblab_cg/__init__.py
MauriceHendrix: Not a bad idea, done

weblab_cg/_chaste_printer.py
MauriceHendrix: Fine we'll call it _custom_functions. It's used in `_print_Function` to be able to print function calls

weblab_cg/_generator_chaste_cg.py
MauriceHendrix: done

weblab_cg/__init__.py
MauriceHendrix: I want to keep to the same terminology as Chaste/PyCML so that we don't confuse everybody. I suppose the CamelCase comees from chast' c++-nes

weblab_cg/_generator.py
```diff
@@ -186,6 +186,7 @@ def derivative_name(deriv):
     # Create parameter information dicts
     parameter_info = []
     for i, parameter in enumerate(parameters):
+        print(parameter)
```
jonc125: Oops!

weblab_cg/__init__.py
```diff
@@ -29,7 +29,20 @@
     load_template,
 )
 
-from ._printer import (     # noqa
+from ._generator_chaste_cg import (   # noqa
```
jonc125: Now we have more stuff here, I'm not convinced by this style of having 'private' modules and then exposing names publicly via `__init__.py`. Why not let users import the modules they actually care about?

Probably don't fix this in this PR though; create an issue for it?

weblab_cg/templates/chaste/normal_model.cpp
```diff
@@ -0,0 +1,157 @@
+//! @file
+//! 
+//! This source file was generated from CellML.
+//! 
+//! Model: {{model_name}}
+//! 
+//! Processed by webalab_cg - CellML Tools in Python: https://github.com/ModellingWebLab/weblab-cg
```
jonc125: 'webalab' ! I'd also remove the 'CellML Tools in Python' bit I think.

weblab_cg/templates/chaste/normal_model.cpp
```diff
@@ -0,0 +1,157 @@
...
+//!     (translators: , webalab_cg: , options: normal)
```
jonc125: Get rid of this. Pycml used it to stick the git revision in generated code; if we're not doing this it's pointless.

weblab_cg/templates/chaste/normal_model.cpp
```diff
@@ -0,0 +1,157 @@
...
+#include <memory>
+#include "Exception.hpp"
+#include "OdeSystemInformation.hpp"
+#include "RegularStimulus.hpp"
+#include "HeartConfig.hpp"
+#include "IsNan.hpp"
+#include "MathsCustomFunctions.hpp"
+    {%- if default_stimulus_equations.membrane_stimulus_current_duration is defined %}
+
+    boost::shared_ptr<RegularStimulus> {{class_name}}FromCellML::UseCellMLDefaultStimulus()
```
jonc125: Could `class_name` not be the full name, rather than just a prefix?

weblab_cg/templates/chaste/normal_model.hpp
```diff
@@ -0,0 +1,83 @@
+#ifndef {% filter upper %}{{class_name}}FROMCELLML_HPP_{% endfilter %}
+#define {% filter upper %}{{class_name}}FROMCELLML_HPP_{% endfilter %}
+
+//! @file
+//! 
+//! This source file was generated from CellML.
+//! 
+//! Model: {{model_name}}
+//! 
+//! Processed by webalab_cg - CellML Tools in Python: https://github.com/ModellingWebLab/weblab-cg
```
jonc125: 'webalab' etc again :)

weblab_cg/tests/chaste_test_utils.py
```diff
@@ -0,0 +1,90 @@
...
+        if lines[i] == '':
+            del lines[i]
+        else:
+            i += 1
+
+    return lines
+
+
+def write_file(file_name, file_contents):
+    """ Write Load a file into a list of lines
```
jonc125: Fix docstring?

weblab_cg/tests/chaste_test_utils.py
```diff
@@ -0,0 +1,90 @@
...
+
+    :param file_name: file name including path
+    :param file_contents: a str with the contents of the file to be written
+    """
+    # Make sure the folder we are writing in exists
+    try:
+        os.makedirs(os.path.dirname(file_name))
```
jonc125: Use `exist_ok=True` instead of catching the exception.

weblab_cg/tests/test_chaste_cg.py
```diff
@@ -0,0 +1,86 @@
...
+
+    def model_types(self):
+        return ['Normal']
+
+    @pytest.fixture(scope="class")
+    def chaste_models(self):
+        """ Load all models"""
+        return load_chaste_models(model_types=self.model_types())
+
+    def test_generate_chatse_models(self, tmp_path, chaste_models):
```
jonc125: 'chatse'

weblab_cg/tests/test_chaste_cg.py
```diff
@@ -0,0 +1,86 @@
...
+        """ Check generation of Normal models against reference"""
+        tmp_path = str(tmp_path)
+        for model in chaste_models:
```
jonc125: Rather than iterating here, you might want to investigate using pytest's test parameterisation to have a separate test case per model. But not in this PR!

weblab_cg/tests/test_chaste_cg.py
```diff
@@ -0,0 +1,86 @@
...
+                expected_cpp = get_file_lines(model['reference_models'][model_type]['expected_cpp_path'])
+
+                # Load generated files
+                generated_hpp = get_file_lines(hhp_gen_file_path)
+                generated_cpp = get_file_lines(cpp_gen_file_path)
+
+                assert expected_hpp == generated_hpp
+                assert expected_cpp == generated_cpp
+
+    def test_generate_dymaic_chatse_model(self, tmp_path):
```
jonc125: `test_generate_dynamic_chaste_model`

weblab_cg/__init__.py
MauriceHendrix: Sounds like something we should discuss, but not for this PR

weblab_cg/_generator.py
MauriceHendrix: Not part of this PR. I'll again update it to the latest master, which should see it gone!

weblab_cg/tests/test_chaste_cg.py
MauriceHendrix: oeps

weblab_cg/templates/chaste/normal_model.cpp
MauriceHendrix: sure

weblab_cg/tests/test_chaste_cg.py
MauriceHendrix: interesting idea I'll look at it at some point.

weblab_cg/templates/chaste/normal_model.cpp
MauriceHendrix: Adding the git revision is a nice idea

mirams: Quite nice to know, for posterity, what commits of weblab_cg / cellmlmanip were involved I suppose

weblab_cg/tests/chaste_test_utils.py
```diff
@@ -73,14 +73,14 @@ def get_file_lines(file_name, remove_comments=False):
...
-    """ Write Load a file into a list of lines
+    """ Write a file into a list of lines
 
     :param file_name: file name including path
     :param file_contents: a str with the contents of the file to be written
     """
     # Make sure the folder we are writing in exists
     try:
-        os.makedirs(os.path.dirname(file_name))
+        os.makedirs(os.path.dirname(file_name), exist_ok=True)
```
jonc125: You don't need the try-except with this change!

Also, I'm not sure the first line of the docstring is correct. Shouldn't it be something like "Write text to the given file"?

jonc125:
I wouldn't worry about trying to hit the codecov target in this PR; it can be improved later.

weblab_cg/templates/chaste/normal_model.cpp
MauriceHendrix: Actually looking at PyCml code it doesn't have git revision on it.

weblab_cg/tests/test_chaste_cg_develop.py
```diff
@@ -0,0 +1,449 @@
...
+            return True
+        eq1 = sympy.sympify(eq1)
+        eq2 = sympy.sympify(eq2)
+
+        e1 = sympy.sympify(self._numbers_with_float_precision(str(eq1)))
+        e2 = sympy.sympify(self._numbers_with_float_precision(str(eq2)))
+        if is_same(e1, e2):
+            return True
+        else:
+            if is_same(eq1, eq2):
```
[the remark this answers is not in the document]
MauriceHendrix: ok

weblab_cg/tests/test_chaste_cg_develop.py
```diff
@@ -0,0 +1,449 @@
...
+        if eq1 == eq2:
+            return True
+        eq1 = sympy.sympify(eq1)
+        eq2 = sympy.sympify(eq2)
+
+        e1 = sympy.sympify(self._numbers_with_float_precision(str(eq1)))
+        e2 = sympy.sympify(self._numbers_with_float_precision(str(eq2)))
+        if is_same(e1, e2):
+            return True
+        else:
```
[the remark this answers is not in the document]
MauriceHendrix: True

weblab_cg/tests/test_chaste_cg_develop.py
```diff
@@ -0,0 +1,449 @@
...
+            if is_same(eq1, eq2):
+                return True
+            if not try_numeric:
+                return False
+            if eq1.free_symbols != eq2.free_symbols:
```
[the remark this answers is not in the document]
MauriceHendrix: yes probably

weblab_cg/tests/test_chaste_cg_develop.py
```diff
@@ -0,0 +1,449 @@
...
+
+        # Remove numbers from lines to check if th text is the same
+        # The escape ascii character is inserted to check placing of number is the same
+        line1 = self._NUM_REGEX.sub(chr(27), line1)
+        line2 = self._NUM_REGEX.sub(chr(27), line2)
+        # Check lines are teh same without the numbers
+        if line1 == line2:
+            # check numbers are the same using sympy
+            if len(num1) == len(num2):
+                for i in range(len(num1)):
```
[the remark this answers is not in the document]
MauriceHendrix: perhaps but I like having i in there so I can put a breakpoint if it doesn't match and use i to pick out thelhs so I can look at the generated files, trying to find it from the equations after sympy has simplified them is not always easy

weblab_cg/tests/test_chaste_cg_develop.py
```diff
@@ -0,0 +1,449 @@
...
+        while substituted_term != eq[1]:
+            eq[1] = substituted_term
+            substituted_term = eq[1].subs(subs_dict)
+        return [eq[0], eq[1]]
+
+    def _resolve_linkers(self, equation_list, subs_dict):
+        """ Process linking variables, remove from the list and update all relevant equations"""
+        # Resolve derivatives
+        remove = []
+        for i in reversed(range(len(equation_list))):
```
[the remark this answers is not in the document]
MauriceHendrix: fair point

weblab_cg/tests/test_chaste_cg_develop.py
```diff
@@ -0,0 +1,449 @@
...
+            if self._is_deriv_decl(equation_list[i][0]):
+                if isinstance(equation_list[i][1], sympy.symbol.Symbol):
+                    # See if there are any linkers
+                    for j in reversed(range(len(equation_list))):
+                        if self._get_var_name(equation_list[j][0]) == str(equation_list[i][1]):
+                            equation_list[i][1] = equation_list[j][1]
+                            remove.append(equation_list[j])
```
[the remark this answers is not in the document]
MauriceHendrix: good point

weblab_cg/tests/test_chaste_cg_develop.py
```diff
@@ -0,0 +1,449 @@
...
+        converter_vars = [eq for eq in equation_list
+                          if (self._is_converter(eq[0]) or isinstance(eq[1], sympy.symbol.Symbol) or
+                              isinstance(eq[1] * -1, sympy.symbol.Symbol)) and not self._is_deriv_decl(eq[0])]
+
+        exclude = []
+        for converter in reversed(converter_vars):
+            # If rhs is symobol, see if there is an equation we can subs in
+            if isinstance(converter[1], sympy.symbol.Symbol) or isinstance(converter[1] * -1, sympy.symbol.Symbol):
+                if self._is_converter(converter[0]):
+                    for i in reversed(range(len(equation_list))):
```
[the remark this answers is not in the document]
MauriceHendrix: ok

weblab_cg/tests/test_chaste_cg_develop.py
```diff
@@ -0,0 +1,449 @@
...
+                    if expected != new_expected:
+                        expected = new_expected
+                        del expected[expected_element_index]
+        assert len(expected) == len(generated)
+
+        # only exception could be name of stimulus current
+        if set([eq[0] for eq in expected]) != set([eq[0] for eq in generated]):
+
+            # Find the 2 differing lhs. There should be 1 and they should have the same rhs:
+            differing_expected = [eq for eq in expected if eq[0] not in [e[0] for e in generated]]
```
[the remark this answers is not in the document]
MauriceHendrix: yeah good point

PR #36, opened 2019-11-13
Web lab generator now fills in the state_var_map.

MichaelClerx:
Emerged as part of https://github.com/ModellingWebLab/weblab-fc/issues/30

### DIFF
--- a/weblab_cg/data/tests/weblab_model.pyx
+++ b/weblab_cg/data/tests/weblab_model.pyx
@@ -163,6 +163,7 @@ cdef class TestModel(CvodeSolver):
         # Mapping from qualified names to state indices; only for states that
         # have a variable name.
         self.state_var_map = {}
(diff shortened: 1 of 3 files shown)


weblab_cg/data/tests/weblab_model.pyx
```diff
@@ -163,6 +163,7 @@ cdef class TestModel(CvodeSolver):
         # Mapping from qualified names to state indices; only for states that
         # have a variable name.
         self.state_var_map = {}
+        self.state_var_map['membrane_voltage'] = 0
```
MauriceHendrix: You could just initialize it in 1 go:
`self.state_var_map = {'membrane_voltage': 0}

MichaelClerx: That's generated code :-D

There can be zero or multiple state vars in that map, and trying to match a template :-)

PR #41, opened 2019-11-27
improving codecov by adding a self excitory model (without stimulus current)

MauriceHendrix:


weblab_cg/data/tests/chaste_reference_models/Normal/demir_model_1994.hpp
```diff
@@ -1,78 +1,78 @@
...
+//! 
+//! <autogenerated>
+
+#include "ChasteSerialization.hpp"
+#include <boost/serialization/base_object.hpp>
+#include "AbstractCardiacCell.hpp"
+
+#include "AbstractStimulusFunction.hpp"
+
+class TestManualdemir_model_1994FromCellML : public AbstractCardiacCell
```
jonc125: I'm not sure about the `Test` prefix on model class names?

weblab_cg/tests/test_chaste_cg_develop.py
```diff
@@ -394,6 +395,23 @@ def _check_equation_list(self, expected, generated):
                     if expected != new_expected:
                         expected = new_expected
                         del expected[expected_element_index]
+
+        # Maybe converters not recognized as converters in the expected?
```
jonc125: I don't understand what this block is doing!

weblab_cg/data/tests/chaste_reference_models/Normal/demir_model_1994.hpp
MauriceHendrix: It's just a name picked by the test suite! We can change it of course let me know if you think it's important. Not for this PR though!

mirams: Yeah, this is very confusing for Chaste code as things starting with Test should actually be tests!

MauriceHendrix: what do you both suggest as prefix? (currently `TestManual`)

weblab_cg/tests/test_chaste_cg_develop.py
MauriceHendrix: it's basically a crude way of checking whether any substitutions have happened above.

weblab_cg/data/tests/chaste_reference_models/Normal/demir_model_1994.hpp
mirams: If will want to be the same as PyCml currently does won't it - `Cell` ? Our is this an intermediate step so we can generate both? If so `CellmlmanipCell` ??

MauriceHendrix: don't confuse these tests with Chaste tests though. This is just a textual comparison on the generated cpp / hpp against a reference, the name is just a variable in the test (we we hook it up the names will be whatever they used to be)

jonc125: Well, if you use the same names as pycml, you don't need to map! Also some (albeit not much) Chaste code might make assumptions about the naming?

MauriceHendrix: what do you mean map?

jonc125:
When checking the generated code matches, you don’t need to convert between weblab-cg naming and pycml naming if they’re the same.

MauriceHendrix:
> When checking the generated code matches, you don’t need to convert between weblab-cg naming and pycml naming if they’re the same.

I'm checking against previously generated code. the develop test checks against the Dynamic version of pycml there the naming is the same. It's easy to change to `cell` if people prefer of course

weblab_cg/tests/test_chaste_cg_develop.py
jonc125: OK, so I kind of get that it's a similar processing to the previous block, but on the newly generated code rather than the reference code from Chaste. However it's not doing exactly the same thing - should it be? Could you add examples of the kind(s) of thing it catches (in both cases actually) to illustrate why it's there?

MauriceHendrix:
> > When checking the generated code matches, you don’t need to convert between weblab-cg naming and pycml naming if they’re the same.
> 
> I'm checking against previously generated code. the develop test checks against the Dynamic version of pycml there the naming is the same. It's easy to change to `cell` if people prefer of course
> 
> actually what looks like naming conversion stuff in the test is names of equation lhs as we couldn't generate the exact names for everything (no access to components and stuff)

to be honest I haven't quite figured this one out 100% yet, it's not doing the same as that didn't quite work. Need to have another good look at this as it has become really complicated.

jonc125:
> to be honest I haven't quite figured this one out 100% yet, it's not doing the same as that didn't quite work. Need to have another good look at this as it has become really complicated.

If you give me some examples of names that don't match I might be able to tell you why!

weblab_cg/data/tests/chaste_reference_models/Normal/demir_model_1994.hpp
jonc125: Sorry, comment below was sent by email and didn't get attached here for some reason. I understand now. Think I'd keep the `Cell` prefix though.

MauriceHendrix:
> > to be honest I haven't quite figured this one out 100% yet, it's not doing the same as that didn't quite work. Need to have another good look at this as it has become really complicated.
> 
> If you give me some examples of names that don't match I might be able to tell you why!

it's not just the names it's also the whole convoluted business of linking things through a multitude of intermediary helper variables

weblab_cg/data/tests/chaste_reference_models/Normal/demir_model_1994.hpp
MauriceHendrix: ok!

weblab_cg/tests/test_chaste_cg_develop.py
MauriceHendrix: changed it a bit as it seems to work both ways but in the basis the issue is going from something like this from PyCml:
```
        const double var_chaste_interface__membrane__i_Stim = GetIntracellularAreaStimulus(var_chaste_interface__environment__time);
        const double var_K__I_stim_converter = var_chaste_interface__membrane__i_Stim; // uA_per_cm2
        const double var_K__chaste_interface__chaste_membrane_capacitance = HeartConfig::Instance()->GetCapacitance(); // uF_per_cm2
        const double var_K__I_stim = var_K__I_stim_converter / var_K__chaste_interface__chaste_membrane_capacitance; // uA_per_uF
```
To this which I get straight from cellmlmanip:
```
const double var_membrane__i_Stim = GetIntracellularAreaStimulus(var_chaste_interface__environment__time)/ HeartConfig::Instance()->GetCapacitance(); // uA_per_uF
```

and then in the case of `EvaluateYDerivatives`
the bit within the else there is (PyCml)
```
        if (mSetVoltageDerivativeToZero)
        {
            d_dt_chaste_interface__membrane__Vm = 0.0;
        }
        else
        {
            const double var_membrane__i_Stim_converter = var_chaste_interface__membrane__i_Stim; // uA_per_cm2
            const double var_membrane__chaste_interface__chaste_membrane_capacitance = HeartConfig::Instance()->GetCapacitance(); // uF_per_cm2
            const double var_membrane__i_Stim = var_membrane__i_Stim_converter / var_membrane__chaste_interface__chaste_membrane_capacitance; // uA_per_uF
.....
            const double var_membrane__kiont = var_membrane__IKr + var_membrane__IKs + var_membrane__IK1 + var_membrane__IKp + ((-2.0) * var_membrane__INaK) + var_membrane__Ito1 + var_membrane__i_Stim; // uA_per_uF
```
while the generated version has:
```
        if (mSetVoltageDerivativeToZero)
        {
            d_dt_chaste_interface__membrane__Vm = 0.0;
        }
        else
        {
.....
            const double var_membrane__kiont = -2.0 * var_INaK__INaK + var_IK1__IK1 + var_IKp__IKp + var_IKr__IKr + var_IKs__IKs + var_Ito1__Ito1 + var_membrane__i_Stim; // uA_per_uF
```

So that bit of code is trying to assess than that's actually fine, since var_membrane__i_Stim was already defined above and defined as the same thing.

jonc125: OK, so essentially you want to be going through extra equations on one side, checking their LHS isn't defined in some other way by the other generator (i.e. they really are extra), then substituting the definition in place of the LHS where it appears in all subsequent equations? That algorithm sounds like it should resolve the mismatch and be the same in both directions?

weblab_cg/tests/test_chaste_cg_develop.py
```diff
@@ -367,6 +366,30 @@ def _resolve_linkers(self, equation_list, subs_dict):
...
+                    for k in range(len(generated)):
+                        if expected[j][0] == generated[k][0]:
+                            generated[k][1] = generated[k][1].subs({self._get_var_name(dif_exp[0]):
+                                                                    dif_exp[1]})
+            if delete:
+                expected_element_index = expected.index(dif_exp)
+                del expected[expected_element_index]
+        return expected, generated
+
+    def _different_eqs(self, expected, generated):
```
jonc125: You only ever read one of the returned pair, so I wouldn't make this calculate both. Instead, call the args say `main_eqs` and `filter_eqs` and calculate just
```
filter_lhss = [eq[0] for eq in filter_eqs]
return [eq for eq in main_eqs if eq[0] not in filter_lhss]
```

weblab_cg/tests/test_chaste_cg_develop.py
jonc125: And might actually mean you don't need the linker handling for the simpler cases?

MauriceHendrix: I have yet to find a real model where no linker handling is needed but in theory yes that's all accurate.
Sadly also some of the lhs are named differently, which sometimes leads to equations that are the same except for some variable names (that's attempted to resolve lower down in the check if the 2 lhs are not the same)

jonc125: Ah, yes, if the LHS names don't match up then you can't just use my straightforward algorithm to cover everything, pity.

Since this is only an interim transition stage anyway, and we have further testing by simulating the generated code, I don't think it's worth spending lots of time trying to refine the process!

MauriceHendrix: Probably not. I've find it pretty helpful though to be able to know what equation causes things not to be the same as for big models the lists can be very long so it's hard to see manually.

weblab_cg/tests/test_chaste_cg_develop.py
```diff
@@ -270,6 +270,7 @@ def _to_equation(self, equation_str):
         """ Do needed substitutions and parse brackets to handle conditionals (cond?exp:exp) """
         # pow->Pow ceil -> ceiling in sympy
         equation_str = equation_str.replace("pow(", 'Pow(').replace("ceil(", 'ceiling(').replace('fabs(', 'Abs(')
+        equation_str = equation_str.replace('M_PI', '3.141592')
```
jonc125: This seems a very low precision version of pi!

weblab_cg/tests/test_chaste_cg_develop.py
MauriceHendrix: ok

weblab_cg/tests/test_chaste_cg_develop.py
MauriceHendrix: actually pi will probably do it

PR #42, opened 2019-11-27
moving abs into _printer.py fixes issue#33

### LINKED ISSUE #33: 'Abs' function unsupported
Generating code for a model that uses 'Abs' tells me that it is not supported.
src/weblab-cg/weblab_cg/_printer.py:203: in _print_Function
   raise ValueError('Unsupported function: ' + str(name))
E   ValueError: Unsupported function: Abs

earm_noble_model_1990.cellml

MauriceHendrix:
fixes #33

### DIFF
--- a/weblab_cg/_chaste_printer.py
+++ b/weblab_cg/_chaste_printer.py
@@ -27,9 +27,6 @@ class ChastePrinter(Printer):
         # Make sure we can output a call to GetIntracellularAreaStimulus
         Printer._custom_functions.add('GetIntracellularAreaStimulus')
 
-    def _print_Abs(self, expr):
-        return 'fabs(' + self._print(expr.args[0]) + ')'
-
     def _print_And(self, expr):
         """ Handles logical And. """
         my_prec = precedence(expr)
--- a/weblab_cg/_printer.py
+++ b/weblab_cg/_printer.py
@@ -107,6 +107,9 @@ class Printer(sympy.printing.printer.Printer):
         else:
             self._derivative_function = derivative_function
 
+    def _print_Abs(self, expr):
+        return self._prefix + 'fabs(' + self._print(expr.args[0]) + ')'
+
     def _bracket(self, expr, parent_precedence):
         """
         Converts ``expr`` to string, and adds parentheses around the result, if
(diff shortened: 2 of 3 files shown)


MichaelClerx:
Couple of comments / questions!

- Why not handle it through the list of functions (`math_functions`) ? It would need a slight rename, but ceil is already handled specially as well. Maybe we could replace it by having a dict of functions _inside_ the printer, that maps sympy names to python or chaste/c++ functions?
- What's the reason for choosing `math.fabs()` over `abs()` ?

MichaelClerx:
@MauriceHendrix Please see above

MauriceHendrix:
> Couple of comments / questions!
> 
> * Why not handle it through the list of functions (`math_functions`) ? It would need a slight rename, but ceil is already handled specially as well. Maybe we could replace it by having a dict of functions _inside_ the printer, that maps sympy names to python or chaste/c++ functions?
> * What's the reason for choosing `math.fabs()` over `abs()` ?

We cant handle it through the list of function as the sympy name is Abs (with capital A) and the function we need to generate isn't called Abs but fabs

math.fabs is for floats math.abs is only for ints

MichaelClerx:
> We cant handle it through the list of function as the sympy name is Abs (with capital A) and the function we need to generate isn't called Abs but fabs

Yes that's why I suggested having a dict instead of a list. And why I pointed out `ceil`, for which we have the same problem but solve it inside `_print_Function`.

> math.fabs is for floats math.abs is only for ints

Just `abs`, not `math.abs`.

MauriceHendrix:
> > We cant handle it through the list of function as the sympy name is Abs (with capital A) and the function we need to generate isn't called Abs but fabs
> 
> Yes that's why I suggested having a dict instead of a list. And why I pointed out `ceil`, for which we have the same problem but solve it inside `_print_Function`.
> 
actually good point we should change that too it's not a great solution!
If you want to have them all in the `_math_functions`  we could store a key, value pairs in the dict for example

> > math.fabs is for floats math.abs is only for ints
> 
> Just `abs`, not `math.abs`.
well `self._prefix + 'fabs'`  and `self._prefix` so happens to be `math.`

MichaelClerx:
> If you want to have them all in the _math_functions we could store a key, value pairs in the dict for example

Yes that's what I'm saying

PR #43, opened 2019-11-28
refactored dealing with function names and removed special case for abs

MauriceHendrix:


weblab_cg/_chaste_printer.py
```diff
@@ -25,7 +25,7 @@ def __init__(self, symbol_function=None, derivative_function=None):
         self._prefix = ''
 
         # Make sure we can output a call to GetIntracellularAreaStimulus
-        Printer._custom_functions.add('GetIntracellularAreaStimulus')
+        Printer._custom_functions.update({'GetIntracellularAreaStimulus': 'GetIntracellularAreaStimulus'})
```
MichaelClerx: Why not `_custom_functions['GetIntracellularAreaStimulus'] = 'Get...' ?

weblab_cg/_printer.py
```diff
@@ -61,34 +61,36 @@ class Printer(sympy.printing.printer.Printer):
...
-        'floor',
-        'log',
-        'log10',
-        'log1p',
-        'log2',
-        'sin',
-        'sinh',
-        'tan',
-        'tanh',
+        'Abs': (lambda x: 'abs' if x.replace('-', '').isnumeric() else 'fabs'),
```
MichaelClerx: ??? Why not just map it to `abs` ? And map the rest to `math.acos` etc.?

weblab_cg/_chaste_printer.py
MauriceHendrix: no particular reason :) I'll change it.

weblab_cg/_printer.py
MauriceHendrix: because it will fail if the number is not an in that's why we could just map it to fabs of course

MichaelClerx: The math.abs and math.fabs methods change the type of the output, which I'm not sure we always want. Seems easier to just use the [built-in `abs` method](https://docs.python.org/3.5/library/functions.html#abs)

MauriceHendrix: what built-in abs method are you talking about?

weblab_cg/_chaste_printer.py
```diff
@@ -17,15 +17,42 @@ class ChastePrinter(Printer):
         A function that converts derivatives to strings.
 
     """
+    _PREFIX = ''
```
MichaelClerx: This can go!

weblab_cg/_printer.py
```diff
@@ -61,34 +61,35 @@ class Printer(sympy.printing.printer.Printer):
...
+        'log2': 'math.log2',
+        'sin': 'math.sin',
+        'sinh': 'math.sinh',
+        'tan': 'math.tan',
+        'tanh': 'math.tanh',
     }
 
-    # List of custom defined functions we are allowed to output
-    _custom_functions = set()
+    # Dictionary of custom defined functions we are allowed to output and their corresponding names to be generated
```
MichaelClerx: This is OK, but could also just add to _math_functions

MauriceHendrix: you didn't want that last time because `GetIntracellularAreaStimulus` isn't a math function :) but we could just call it functions

MichaelClerx: Yeah with the prefixes gone that makes sense. Thanks!

MauriceHendrix: ok I also moved the default dict into the _weblab_printer, just to hopefully make more clear that we'd expect it to be different for each one. let me know what you think

weblab_cg/_printer.py
```diff
@@ -61,40 +61,14 @@ class Printer(sympy.printing.printer.Printer):
...
-    _custom_functions = set()
+    # Dictionary of functions we can handle and their corresponding names to be generated
+    _function_names = {}
 
     def __init__(self, symbol_function=None, derivative_function=None):
         super(Printer, self).__init__(None)
 
         # Prefix for functions
-        self._prefix = 'math.'
+        self._prefix = ''
```
MichaelClerx: This still needs to be "math." to make it work for sqrt etc.

An alternative would be to
1. Remove self._prefix
2. Add "sqrt" and co into the list of functions
3. Make functions that now use `self._prefix` lookup the correct name for "sqrt" in the list of functions

PR #44, opened 2019-12-02
Used pytest.parametrize in test_chaste_cg to get a bit more response when testing.

MichaelClerx:
Hey @MauriceHendrix 

The test_chaste_cg test takes quite a while, and I didn't see any output so was worried it was broken. Turns out it's just a long test!

Did you know about `pytest.parametrize`? It lets you give a list of arguments to test with, e.g. a list of models. This has the advantage that each model becomes a `.` on the screen when testing so you get a bit more output, but also when a model fails it will tell you which model.

Just in case you like it :D

### DIFF
--- a/weblab_cg/tests/test_chaste_cg.py
+++ b/weblab_cg/tests/test_chaste_cg.py
@@ -12,46 +12,48 @@ LOGGER = logging.getLogger(__name__)
 LOGGER.setLevel(logging.DEBUG)
 
 
+def model_types():
+    return ['Normal']
+
+
+def chaste_models():
+    """ Load all models"""
+    return load_chaste_models(model_types=model_types())
+
+
 class TestChasteCG(object):
     """ Tests weblab_cg against reference models generated with weblab_cg and tested in chaste.
 
     TODO: unit conversion vvia cellmlmanip once implemented"""
     _TIMESTAMP_REGEX = re.compile(r'(//! on .*\n)')
 
-    def model_types(self):
-        return ['Normal']
-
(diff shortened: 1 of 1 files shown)


MichaelClerx:
You still spend a long time waiting before the tests can begin with this method, but maybe that can be fixed by splitting `load_chaste_models` into `list_chaste_models()` and `load_chaste_model(path)` ?

MauriceHendrix:
> Hey @MauriceHendrix
> 
> The test_chaste_cg test takes quite a while, and I didn't see any output so was worried it was broken. Turns out it's just a long test!
> 
> Did you know about `pytest.parametrize`? It lets you give a list of arguments to test with, e.g. a list of models. This has the advantage that each model becomes a `.` on the screen when testing so you get a bit more output, but also when a model fails it will tell you which model.
> 
> Just in case you like it :D

Yes @jonc125 told me about it it was on my todo list actually (but I hadn't started at it). 
I haven't found a way to write output that you can see during the test. When i write to the logs or stdout/stderror it all just shows after the test fails. During the test or after succesful completion we see nothing.

Plan for the future of course is to reduce the set of models tested on a daily basis and/or maybe to make some artificial once that have what we need.

MichaelClerx:
> I haven't found a way to write output that you can see during the test. When i write to the logs or stdout/stderror it all just shows after the test fails. During the test or after succesful completion we see nothing.

That really annoys me too! I think you can add `-s` or something to see it, but then it's still hard to spot sometimes

PR #45, opened 2019-12-03
More test models

MauriceHendrix:
## Description

It turns out my understanding about how capacitance is used in converting stimulus currents was not quite accurate. Units can dimensionally equivalent to be uA_per_cm2, uA or uA_per_uF.
Initially I thought that both membrane_capacitance and HeartConfig::Instance()->GetCapacitance() were used when the units aren't (demensionally equivalent to) uA_per_cm2.
However it seems that for uA_per_uF the membrane_capacitance is not used. I hadn't noticed this before as the other models I tried with these units have no membrane_capacitance variable so it wasn't used in those.

I also changed the way this is recorded, into a dictionary so it's easier to extend and the method doing the conversion doesn't have to check for specific unit cases.  When adding units to the model, the dictionary gets copied to a class variable and populated with model units for eas of use.

I moved `_HEARTCONFIG_GETCAPACITANCE = 'HeartConfig::Instance()->GetCapacitance()` out at the top of the class declaration, so we don't have to type it multiple times and it would be easier to change.

## Motivation and Context

I want to be able to transform models, and at least the current pycml known working ones. Also keep correcting mistakes in the logic.

## Types of changes

## Checklist:

Tried to anyway!
##Testing


weblab_cg/_generator_chaste_cg.py
```diff
@@ -537,10 +537,9 @@ def _format_derivative_equations(self):
                 divisor_unit = self._get_desired_units(d_eqs[i].lhs.args[1][0])
                 units = dividend_unit / divisor_unit
                 units_from = self._model.units.summarise_units(d_eqs[i].lhs)
-                if units.dimensionality == units_from.dimensionality:
-                    factor = self._model.units.get_conversion_factor(units, from_unit=units_from)
-                else:
-                    factor = 1.0
+                factor = self._model.units.get_conversion_factor(units, from_unit=units_from) \
```
jonc125: That's a bit of a cheat way of fixing coverage ;)

MauriceHendrix: I Know, however I know the else case will be covered for the next model I'm trying. i just decided when adding that model I should really split up PRs, and I'd rather not take out the check and then again spent time on wondering why the next model isn't working. So the next model will use the else case, promise!

MauriceHendrix:
> I'm afraid I'm not going to be able to look through the logic of this properly in the near future. If the tests are passing I'm happy to trust that it's correct. We'll probably want to refactor the units handling in the future anyway once the more general functionality for Web Lab is added!

yeah fair enough I'm increasingly trying to move as much of `if variable is in these units do X` type code out of the main functions and into somewhere where it'll be easier to extend and change in the future and hopefully make the main functions easier to follow (slowly)

MichaelClerx:
It's not a style/philosophical point! But when I change a comma somewhere, I want to wait ~10s for the unit tests to complete so I can commit, not minutes!

MauriceHendrix:
> Do we need this many test files? Can we not just pick a specific model that has the feature/difficulty we're trying to test?

We could but an important point of the tests is to test we haven't broken old working feature right? So for example this PR includes a change in the way we deal with capacitance, which affects the previous models too. 
Maybe there's some way of specifying a bunch f tests to be skipped for you?

MichaelClerx:
But why is it not good enough to have 1 model that requires using C, and one that doesn't. Why lots of each?

MauriceHendrix:
> But why is it not good enough to have 1 model that requires using C, and one that doesn't. Why lots of each?

C?
There'sat least 3 different options for membrane_current units , 2 for membrane_capacitance 
and then you also have self excitory models without any of that, statically linked vs dynamic models 

So that's at least 8 options already.

MichaelClerx:
Fair enough. What are the 2 options for capacitance?

MauriceHendrix:
> Fair enough. What are the 2 options for capacitance?

`uF`, `uF_per_mm2` (and not having a capacitance tagged so that's kinda 3)

jonc125:
On the test quantity front, eventually we'll have much shorter tests because they won't have to do all the work to compare with pycml, and we'll select just a few to run on every commit, with longer ones on lots of models running nightly or weekly. (Chaste tests most pycml functionality only nightly with `TestPycmlLong`.)

If testing locally before commit, it's certainly OK to skip tests you're fairly sure won't be impacted. There are various pytest options you can use for this, notably passing just the test file names you want to run! We could also look at [custom marks](https://docs.pytest.org/en/latest/mark.html#mark) to make this easier (so you could do e.g. `pytest -m "not chaste"`).

MauriceHendrix:
> On the test quantity front, eventually we'll have much shorter tests because they won't have to do all the work to compare with pycml, and we'll select just a few to run on every commit, with longer ones on lots of models running nightly or weekly. (Chaste tests most pycml functionality only nightly with `TestPycmlLong`.)
> 
> If testing locally before commit, it's certainly OK to skip tests you're fairly sure won't be impacted. There are various pytest options you can use for this, notably passing just the test file names you want to run! We could also look at [custom marks](https://docs.pytest.org/en/latest/mark.html#mark) to make this easier (so you could do e.g. `pytest -m "not chaste"`).

see #49

MichaelClerx:
Cool!
Which model uses `uF_per_mm2`? That's usually taken to be a (cell-independent) constant

MauriceHendrix:
beeler_reuter_model_1977 and luo_rudy_1994

PR #50, opened 2019-12-09
Removed sort also for other methods that get derivative eqs. Reintroduced excluding of modifiable parameters

MauriceHendrix:
## Description

I released that I only removed sorting of equations for derivatives in 1 place, so i removed in in the other places too (you can get the eqs, the eqs without the voltage ones, or only voltage ones).

These 2 sets (eqs without the voltage ones and only voltage ones) weren't directly being output, but used as reference in the formatting function to set the in_voltage flag. The sorts aren't needed though so I took them out to make the code simpler.

I also noticed when taking the sort out of the _get_derivative_equations I took out everything also the bit that prevents equations where the lhs is a modifiable parameter from appearing. This has been corrected.

## Motivation and Context

I was getting C++ compile errors because I was generating things like:
`const double mParameters[0] = 1.0`
Caused by having taken out the code to filter out equas where the lhs is a modifiable parameter. 

## Types of changes

## Checklist:

##Testing


weblab_cg/tests/test_chaste_cg_develop.py
```diff
@@ -33,8 +33,6 @@ def chaste_models(self):
         return load_chaste_models(model_types=self.model_types(),
                                   ref_path_prefix=['chaste_reference_models', 'develop'], class_name_prefix='Dynamic')
 
-    @pytest.mark.skip(reason="This test is a development tool")
```
jonc125: The marks have gone here. From dd4873f it seems that was unintentional?

MauriceHendrix: I forgot the chaste mark thanks! I deliberately removed the other one as manually changing it was driving me nuts, when there are no reference models it'll be skipped anyway.

PR #53, opened 2019-12-11
Additional models

MauriceHendrix:
## Description
I have made a few fixes for issues higlighted by some of the models. They are part interdependant so hence the single PR.
- Added check to throw nicer error when trying to apply conversion rule to stimulus current without having a membrane_capacitance. Also added test for the error.
_use_get_intracellular_calcium_concentration has moved to the generate_chaste_code as it is a very simple condition.
- added code to deal with state_vars that have `range-low` and `range-high` annotation, also added `get_annotated_var_name`
- updated calculation of the sign for `GetIntracellularAreaStimulus`
`generate_chaste_code` has moved.
- All but 2 of the refernce models have been added to the test, most in additional_models (these converted out of the box and are skipped)

## Motivation and Context

- Error on missing membrane_capacitance was not helpful.
- _use_get_intracellular_calcium_concentration was not needed, the condition is simple enough.
- chaste code generated uses range-low and range-high annotation in `VerifyStateVariables`, `get_annotated_var_name` can get the more popper name to use in the error message in the generated code.
- The sign fr `GetIntracellularAreaStimulus` was incorrect for ne of the models, as the code was not properly searching for variable definitions.
- Moving `generate_chaste_code` to the end means all private methods are before the public one.
- We want to verify we can convert all models.
## Types of changes

## Checklist:

##Testing


MichaelClerx:
I don't think I can review this. Maybe @mirams ?

MauriceHendrix:
> I don't think I can review this. Maybe @mirams ?

@mirams  @jonc125  @MichaelClerx how are we going to do this going forward?

MichaelClerx:
Gary and Jonathan are the chaste experts! I can't tell if these are right.

mirams:
If its code only Maurice is working on and worried about I'd suggest he's allowed to merge it in himself!

MichaelClerx:
Another good reason to implement #54 ;-)

PR #57, opened 2019-12-16
Update setup.py

MauriceHendrix:
## Description

Pinned sympy version to 1,4 in setup.py

## Motivation and Context

The different sympy version (1.5) the CI was pulling in was causing my tests to fail in CI but pass locally

## Types of changes

## Checklist:

No new comments added

##Testing

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -41,6 +41,7 @@ setup(
     install_requires=[
         # cellmlmanip                # Add this in when cellmlmanip is ready
         'Jinja2>=2.10',
+        'sympy==1.4',
     ],
     extras_require={
         'test': [


MichaelClerx:
Shouldn't we be fixing cellmlmanip to work with sympy 1.5 instead?

Also we had some discussion before about pinning versions in setup.py (I'm having some trouble finding links to it!) and the conclusion was that we can pin things in `dev.txt` and similar (so that devs use the same version, which @jonc125 says is a good idea) but not in `setup.py` (because its a library so needs to work in several environments). 

Specifying a minimum version in `setup.py` is ok, or if there's a version of a dependency with a known bug you can disallow that version explicitly. But in this case it looks like sympy behaviour has changed so we need to update cellmlmanip to work with the new behaviour and bump the required version up to 1.5?

MauriceHendrix:
> Shouldn't we be fixing cellmlmanip to work with sympy 1.5 instead?
> 
> Also we had some discussion before about pinning versions in setup.py (I'm having some trouble finding links to it!) and the conclusion was that we can pin things in `dev.txt` and similar (so that devs use the same version, which @jonc125 says is a good idea) but not in `setup.py` (because its a library so needs to work in several environments).
> 
> Specifying a minimum version in `setup.py` is ok, or if there's a version of a dependency with a known bug you can disallow that version explicitly. But in this case it looks like sympy behaviour has changed so we need to update cellmlmanip to work with the new behaviour and bump the required version up to 1.5?



> Shouldn't we be fixing cellmlmanip to work with sympy 1.5 instead?
> 
> Also we had some discussion before about pinning versions in setup.py (I'm having some trouble finding links to it!) and the conclusion was that we can pin things in `dev.txt` and similar (so that devs use the same version, which @jonc125 says is a good idea) but not in `setup.py` (because its a library so needs to work in several environments).
> 
> Specifying a minimum version in `setup.py` is ok, or if there's a version of a dependency with a known bug you can disallow that version explicitly. But in this case it looks like sympy behaviour has changed so we need to update cellmlmanip to work with the new behaviour and bump the required version up to 1.5?

No it needs to be the same version otherwise we get this situation where tests pass locally but not in CI. If you want we can do >=1.4 in the dev requirements instead though

MichaelClerx:
> otherwise we get this situation where tests pass locally but not in CI.

That's a feature! You test locally in whatever, and then ideally CI tests in all other supported configurations (or some sensible subset). At the moment we're not doing this very thoroughly, but having setup.py say something like `>=1.4` will make it pull in new versions and alert us to issues with newer versions. If we had everything pinned it would hide this issue, not solve it

MauriceHendrix:
> > otherwise we get this situation where tests pass locally but not in CI.
> 
> That's a feature! You test locally in whatever, and then ideally CI tests in all other supported configurations (or some sensible subset). At the moment we're not doing this very thoroughly, but having setup.py say something like `>=1.4` will make it pull in new versions and alert us to issues with newer versions. If we had everything pinned it would hide this issue, not solve it

I understand you want to cater for new versions, but this is not a good way to do it in my opinion.

jonc125:
From Slack:
> What I normally do on projects, is have CI run tests at least twice: once with the pinned dev versions (so if it works locally it'll pass there) and also as an 'allowed failure' with latest versions. So you get alerted if a new version has just appeared that breaks things, but don't have to fix it immediately to merge a feature PR.
> If you're officially supporting multiple versions you can also run with those, and decide on priority as to whether they're allowed failures or not.

For this PR, I'd suggest either just merge (!) or fix by updating the Travis and AppVeyor config in weblab-cg to use the pinned versions, then in a second PR update the configs again to test both pinned & latest.

jonc125:
@MichaelClerx says Cellmlmanip is tested with both so can use its config as an example.

PR #59, opened 2019-12-17
found a solution for our float problem using strip_units=False

MauriceHendrix:
## Description

Changed to using strip_units=False in all get_equations_for calls and updated the printer and stimulus sign calculation to cope with this change.

## Motivation and Context

We were having issues where the simplification of sympy went to far.
Fix / workaround for [#189](https://github.com/ModellingWebLab/cellmlmanip/issues/189) (allows me to generate correct code)

## Types of changes

## Checklist:

##Testing


MauriceHendrix:
> Cunning! Seems a shame to have to do it, but let's go with this for now, since it works :)

Totally, hopefully we'll find a better solution in sympy at some point.

PR #63, opened 2020-01-14
Unit conversion via cellmlmanip

MauriceHendrix:
## Description

This change uses cellmlmanip unit conversion for _time_, _voltage_ and _cytosolic_calcium_concentration_ a few things had to be moved around to facilitate this. Conversion of stimulus currents and _i_ionic_ using _capacitance_ is still done in _chaste_cg_

## Motivation and Context

It's useful to have as much of the unit conversion logic in 1 place, for maintanance reasons. So it makes sense to use cellmlmanip unit conversion where possible

## Types of changes

## Checklist:

##Testing


weblab_cg/_generator_chaste_cg.py
```diff
@@ -196,6 +197,42 @@ def _add_units(self):
...
+            desired_units = self._model.units.ureg.millimolar
+
+            # Convert if necessary
+            try:
+                return self._model.convert_variable(cytosolic_calcium_concentration,
+                                                    desired_units, DataDirectionFlow.INPUT)
+            except DimensionalityError:
+                # units can't be converted to millimolar (e.g. could be dimensionless)
+                return cytosolic_calcium_concentration
+        except KeyError:
```
jonc125: I'd move this `except` up before you define `desired_units`. Then you don't need to nest another try-except inside the `try` clause, it can come out a level.

weblab_cg/_generator_chaste_cg.py
```diff
@@ -601,131 +662,60 @@ def _format_default_stimulus(self):
...
-                         'var_name': self._printer.doprint(self._time_variable)}
-        system_info = \
-            [{'name': self._model.get_ontology_terms_by_symbol(var, self._OXMETA)[0]
-              if self._model.has_ontology_annotation(var, self._OXMETA) else self._name_printer.doprint(var),
-              'initial_value': str(self._model.get_initial_value(var))
-                if var not in self._state_var_conversion_factors
-                else str(self._state_var_conversion_factors[var] * self._model.get_initial_value(var)),
-              'units': self._get_desired_units(var)}
-             for var in self._state_vars]
+        return {'name': name,
```
jonc125: Could you just use `var_display_name` to calculate `name` rather than repeating checking ontology logic etc? (Perhaps not in this PR!)

weblab_cg/_generator_chaste_cg.py
```diff
@@ -601,131 +662,60 @@ def _format_default_stimulus(self):
...
+                'units': self._model.units.summarise_units(self._time_variable),
+                'system_name': self._model.name,
+                'var_name': self._printer.doprint(self._time_variable)}
 
+    def _format_system_info(self):
+        """ Format general ode system info for chaste output"""
+        return [{'name': self._model.get_ontology_terms_by_symbol(var, self._OXMETA)[0]
```
jonc125: There's very similar logic here, for instance.

weblab_cg/data/tests/chaste_reference_models/Normal/additional_models/Shannon2004.cpp
```diff
@@ -703,37 +703,37 @@ void OdeSystemInformation<CellShannon2004FromCellML>::Initialise(void)
     this->mInitialConditions.push_back(0.001405627);
     
     // rY[5]:
-    this->mVariableNames.push_back("ar_IKr_Xr_gate__Xr");
+    this->mVariableNames.push_back("var_IKr_Xr_gate__Xr");
```
jonc125: Hah, oops :)

weblab_cg/_generator_chaste_cg.py
MauriceHendrix: good spot!

weblab_cg/_generator_chaste_cg.py
MauriceHendrix: another good spot. It's really just because the _get_var_display_name was introduced much later that I didn't spot it.

PR #70, opened 2020-01-28
Switch strip units off

MauriceHendrix:
## Description

- Instead of using `strip_units=False` in `self._model.get_equations_for` we now make cellmlmanip encode exp as _exp and print exp in the printer
- created my own _get_equations_for that exclude equations where the lhs is a parameter
- Fixed get_initial_value issue for modifiable parameters 

## Motivation and Context

- Allows using more of sympy's simplification. no more unnecessary `num * num` etc.
- in a lot of places I was excluding such equations so introduced a function for it
- The way it changed in cellmlmanip has changed so get_initial_value no longer worked. Only just noticed that

## Types of changes

## Checklist:

## Testing


weblab_cg/_generator_chaste_cg.py
```diff
@@ -123,13 +126,11 @@ def __init__(self, model, file_name, **kwargs):
         self._ionic_derivs = self._get_ionic_derivs()
         self._equations_for_ionic_vars, self._current_unit_and_capacitance, self._membrane_stimulus_current_factor = \
             self._get_equations_for_ionic_vars()
+        self._equations_for_ionic_vars = self._equations_for_ionic_vars
```
jonc125: no-op!

weblab_cg/_generator_chaste_cg.py
```diff
@@ -193,20 +199,26 @@ def _add_units(self):
     def _get_time_variable(self):
         time_variable = self._model.get_free_variable_symbol()
         desired_units = self._model.units.ureg.millisecond
-        assert self._model.units.summarise_units(time_variable).dimensionality == desired_units.dimensionality, \
-            'Incorrect definition of time variable (units of time needs to be demensionally equivalent to second)'
-        # If the variable is in units that can be converted to millimolar, perform conversion
-        return self._model.convert_variable(time_variable, desired_units, DataDirectionFlow.INPUT)
+        try:
+            # If the variable is in units that can be converted to millimolar, perform conversion
```
jonc125: s/millimolar/millisecond/ !

PR #71, opened 2020-01-29
Partial eval

MauriceHendrix:
## Description

Added partial evaluation in preparation for lookup tables (for optimized  models)
Had to move about the way the stimulus is pricessed

## Motivation and Context

This is needed for for lookup tables (for optimized  models)

## Types of changes

## Checklist:

## Testing


weblab_cg/_generator_chaste_cg.py
```diff
@@ -179,6 +192,30 @@ def _get_var_display_name(self, var):
             display_name = var.cmeta_id
         return display_name.replace('$', '__').replace('var_chaste_interface_', '', 1).replace('var_', '', 1)
 
+    def _partial_eval(self, equations, required_lhs):
```
jonc125: One thing I learnt through doing pycml was to move these model analysis/transformation steps into their own class/module, rather than embedded in other parts of the code. So I'd be inclined to create a new module to hold this - it's just a standalone function after all! Not sure what you'd call it; guess it depends on whether lookup tables goes in the same file, and where things like rush-larsen end up going. You might want a sub-package for all analysis/transformations.

weblab_cg/_generator_chaste_cg.py
```diff
@@ -50,15 +51,14 @@ class ChasteModel(object):
                    'membrane_stimulus_current_end': [{'units': 'millisecond'}],
                    'membrane_capacitance': [{'units': 'uF'}, {'units': 'uF_per_mm2'}]}
 
-    _STIM_RHS_MULTIPLIER = {'membrane_stimulus_current_amplitude':
-                            {'uA_per_uF': sp.Function(_HEARTCONFIG_GETCAPACITANCE)(),
-                             'uA': sp.Function(_HEARTCONFIG_GETCAPACITANCE)()}}
-
     _STIM_CONVERSION_RULES_ERR_CHECK = \
```
jonc125: Might be useful to have some comments explaining how these are used?

weblab_cg/_generator_chaste_cg.py
```diff
@@ -315,16 +352,55 @@ def _get_membrane_capacitance(self):
...
+                for units_to_try in [unit_dict['units'] for unit_dict in self._stim_units[key]]:
+                    if units_to_try.dimensionality == current_units.dimensionality:
+                        units = units_to_try
+                        factor = self._model.units.get_conversion_factor(units, from_unit=current_units)
+                        if factor != 1.0:
+                            warning = 'converting ' + str(key) + ' from ' + str(current_units) + ' to ' + str(units)
+                            self._logger.info(warning)
+                        # apply convrsion rule if we have one
+                        if key in self._STIM_CONVERSION_RULES and str(units) in self._STIM_CONVERSION_RULES[key]:
+                            warning = 'Converting ' + str(key) + ' from ' + str(units) + ' into chase units'
```
jonc125: chase -> Chaste :)

weblab_cg/_generator_chaste_cg.py
```diff
@@ -179,6 +192,30 @@ def _get_var_display_name(self, var):
...
 
+    def _partial_eval(self, equations, required_lhs):
+        # only do partial eval if pe option is set
+        if not self._pe:
+            return equations
+
+        evaluated_eqs = list(reversed(equations))
+        # count usage of variables on rhs of equations
+        usage_count = dict()
+        for i in range(len(evaluated_eqs) - 1, - 1, - 1):
```
jonc125: Better: `for eq in equations:`

weblab_cg/_generator_chaste_cg.py
```diff
@@ -179,6 +192,30 @@ def _get_var_display_name(self, var):
...
+            usage_count[evaluated_eqs[i].lhs] = 0
+            for var in evaluated_eqs[i].rhs.atoms(VariableDummy):
+                usage_count[var] = usage_count[var] + 1 if var in usage_count else 1
```
jonc125: You may find the `dict.setdefault` method handy here.

weblab_cg/_generator_chaste_cg.py
```diff
@@ -179,6 +192,30 @@ def _get_var_display_name(self, var):
...
+
+        # subs in all constants and expressions only used once
+        subs_dict = {}
+        for i in range(len(evaluated_eqs) - 1, - 1, - 1):
```
jonc125: Rather than going through this list backwards, you could just append to the result list and iterate forwards through the original:
```python
evaluated_equations = []
for eq in equations:
    new_eq = eq.subs(subs_dict)
    if new_eq.lhs not in # etc.
        subs_dict[new_eq.lhs] = new_eq.rhs
    else:
        evaluated_equations.append(new_eq)
```

weblab_cg/_generator_chaste_cg.py
MauriceHendrix: didn't know you could do that!

weblab_cg/_generator_chaste_cg.py
MauriceHendrix: I didn't want to waste extra memory although to be fair` evaluated_eqs = list(reversed(equations))` kind of does so anyway

jonc125: Indeed! And appending to a list is much cheaper (amortised) than deleting from the middle of one.

weblab_cg/_generator_chaste_cg.py
MauriceHendrix: Ok good point

weblab_cg/_generator_chaste_cg.py
MauriceHendrix: Actually good point I realized I forgot to do it, but it was exactly my intention when I created the OO structure, now addressed.

weblab_cg/_command_line_script.py
```diff
@@ -9,24 +9,26 @@
...
+    # Link names to classes for converting code
+    translators = {'Chaste': cg.NormalChasteModel, 'ChasteOpt': cg.OptChasteModel}
+    # Store extensions we can use and how to use them, based on extension of given outfile
+    extension_lookup = {'Chaste': dict(), 'ChasteOpt': dict()}
     for key in ('.cpp', '.hpp', '.cellml', ''):
         extension_lookup['Chaste'][key] = ['.hpp', '.cpp']
+        extension_lookup['ChasteOpt'][key] = ['.hpp', '.cpp']
     for key in ('.c', '.h'):
         extension_lookup['Chaste'][key] = ['.h', '.c']
+        extension_lookup['ChasteOpt'][key] = ['.hpp', '.cpp']
```
jonc125: Shouldn't this be .h, .c?

weblab_cg/_generator_chaste_cg.py+
```diff
@@ -20,12 +20,13 @@ class ChasteModel(object):
     # TODO: script to call generator from command line
```
jonc125: This file appears to have been renamed with a + in the extension??

weblab_cg/_generator_chaste_cg.py+
```diff
@@ -20,12 +20,13 @@ class ChasteModel(object):
...
     # TODO: Model types Opt, Analytic_j, Numerical_j, BE
 
-    _MEMBRANE_VOLTAGE_INDEX = 0
-    _CYTOSOLIC_CALCIUM_CONCENTRATION_INDEX = 1
-    _OXMETA = 'https://chaste.comlab.ox.ac.uk/cellml/ns/oxford-metadata#'
-    _PYCMLMETA = 'https://chaste.comlab.ox.ac.uk/cellml/ns/pycml#'
+    _MEMBRANE_VOLTAGE_INDEX = 0  # default index for voltage in state vector
+    _CYTOSOLIC_CALCIUM_CONCENTRATION_INDEX = 1  # default index for voltage in state vector
```
jonc125: This one isn't voltage :)

weblab_cg/_command_line_script.py
MauriceHendrix: uups thanks for noticing!

weblab_cg/_generator_chaste_cg.py+
MauriceHendrix: Thanks ... a bit too trigger happy with the copy-paste

PR #74, opened 2020-02-03
moved to own initial_value function

MauriceHendrix:
## Description

moved to own initial_value function

## Motivation and Context

cellmlmanip version no longer works for non-state variables and may be removed. I found it convenient to have my own.
see also: https://github.com/ModellingWebLab/cellmlmanip/issues/201

## Types of changes

## Checklist:

## Testing


weblab_cg/chaste_model.py
```diff
@@ -173,6 +173,17 @@ def _get_equations_for(self, symbols, recurse=True):
...
+    def _get_initial_value(self, var):
+        """Returns the initial value of a variable if it has one, none otherwise"""
+        # state vars have an initial value parameter defined
+        if var in self._state_vars:
+            return getattr(var, 'initial_value', None)
+        eqs = self._model.get_equations_for([var])
+        # If there is a defining equation, there should be just 1 equation and it should be of the form var = value
+        if len(eqs) == 1 and isinstance(eqs[0].rhs, sp.numbers.Float):
+            initial_value = eqs[0].rhs
+        return initial_value
```
jonc125: You should probably raise a sensible error in an else clause, otherwise you'll get a 'local variable referenced before definition' here.

PR #80, opened 2020-02-03
Updated travis URL

MichaelClerx:
From .org to .com

Second version, to see if the webhooks for .org are gone


MauriceHendrix:
Updated travis settings for new repo name

PR #92, opened 2020-02-25
catching up with cellmlmanip

MauriceHendrix:
## Description

Catching up with cellmlmanip name changes and the fact cemata_id is no longer publicly writable

## Motivation and Context

changes are needed to keep chaste_codegen working as the api signature of cellmlmanip has changed
also renamed some test files. As the repo no longer contains other code it seemed unnecessary to refer to the tests as chaste_codegen specific tests

## Types of changes

## Checklist:

## Testing
Also ran & past chaste tests for normal & opt models. Cvode models don't pass but they didn't previously either at this stage (will be fixed in next pr).
I had to update the cellml files as a few minor updates have happened to the ones used in the chaste test (or rather reverts to the version pycml chaste tests use)


chaste_codegen/data/tests/cellml/hilgemann_noble_model_1987.cellml
```diff
@@ -75,16 +75,10 @@ http://www.CellML.org/
...
    </units>
    <component name="environment">
-      <variable name="time" public_interface="out" units="second" cmeta:id="time">
-         <rdf:RDF xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#" xmlns:bqbiol="http://biomodels.net/biology-qualifiers/">
-            <rdf:Description rdf:about="#time">
-               <bqbiol:is rdf:resource="https://chaste.comlab.ox.ac.uk/cellml/ns/oxford-metadata#time"/>
-            </rdf:Description>
-         </rdf:RDF>
-      </variable>
+      <variable name="time" units="second" public_interface="out"/>
```
jonc125: Why have all the CellML files changed too?

MauriceHendrix: because I've reverted to the original cellml files just with metadata tags added.
What I had done initially is just get the new versions, which was fine for normal conversions but PyCml didn't generate jacobians for a number of the new cellml files, so I reverted in order to compare apples with apples.

PR #93, opened 2020-02-26
resolve inconsistent behaviout with cronjob tests

MauriceHendrix:
## Description

Fixes interference between models.

## Motivation and Context

When running cron job tests some models were failing, while they were passing in isolation. This test fixes the interference.

## Types of changes

## Checklist:

n/a

## Testing
also tested by (manually) running  cronjob tests which is where the issue was highlighted

### DIFF
--- a/chaste_codegen/chaste_model.py
+++ b/chaste_codegen/chaste_model.py
@@ -13,11 +13,8 @@ class ChasteModel(object):
     It also holds relevant formatted equations and derivatives.
     Please Note: this calass cannot generate chaste code directly, instead use a subclass of the model type
     """
-    # TODO: implement modifiable-parameter bit of expose-annotated-variables (similar to derived quantities)
     # TODO: implement and check options see https://chaste.cs.ox.ac.uk/trac/wiki/ChasteGuides/CodeGenerationFromCellML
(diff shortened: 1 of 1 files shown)


chaste_codegen/chaste_model.py
```diff
@@ -415,7 +412,7 @@ def _get_equations_for_ionic_vars(self):
                                             == self._units.get_unit(unit_cap['units']).dimensionality
                                             and eq.lhs not in self._ionic_derivs]
                 equations = [eq.lhs for eq in equations]
-            desired_units_and_capacitance = unit_cap
+            desired_units_and_capacitance = unit_cap.copy()
```
MichaelClerx: This is interesting!
Is `unit_cap` a pint.Unit?

MauriceHendrix: no it's a dictionary with a pint unit and some other information to decide how to handle units dimensionally equivalent to the unit. The things is that I change `desired_units_and_capacitance['units']` later down and that unit_cap points to a class level variable hence the interference

MichaelClerx: Ah. Thanks!

PR #103, opened 2020-03-02
trying to split up cronjob

MauriceHendrix:
## Description

reduced number of model tested in cronjob (all models but not every variant is tested)

## Motivation and Context

the cronjob was timing out

## Types of changes

chanage to test setup

## Checklist:

n/a

## Testing
tested by setting the wrong flag. It seems to works


chaste_codegen/tests/test_cronjob.py
```diff
@@ -0,0 +1,53 @@
...
+
+chaste_all_normal_models = get_models(ref_folder='cronjob_reference_models', type='Normal')
+chaste_all_opt_models = get_models(ref_folder='cronjob_reference_models', type='Opt')
+chaste_all_cvode_models = get_models(ref_folder='cronjob_reference_models', type='Cvode')
+chaste_all_cvode_models_jacobian = get_models(ref_folder='cronjob_reference_models', type='Cvode_with_jacobian')
+
+
+@pytest.mark.cronjob
+@pytest.mark.parametrize(('model'), chaste_all_normal_models)
+def test_Normal_cronjob(tmp_path, model, request):
```
jonc125: I was thinking that either you'd want different marks for each test here, or change your travis cronjob setup so you have 4 (say) jobs, adding arguments '-k test_Normal', '-k test_Opt', etc. But splitting in this way might be enough for now - it depends how long these take!

MauriceHendrix: I tried it doesn't seem to help, seems to go off overall time unless I'm missing something

PR #106, opened 2020-03-04
split templates in prep for backwards euler

MauriceHendrix:
## Description

Split templates and included partial templates
Also needed to update reference files as I changed the order of 1 of the includes for conveiance

## Motivation and Context

This is to make it easier to make the backwards Euler template and easier to maintain hopefully

## Types of changes

## Checklist:

n/a

## Testing
new references also tested in chaste


chaste_codegen/templates/cvode_model.cpp
```diff
@@ -1,241 +1,11 @@
...
-    AbstractCardiacCellInterface* MakeCardiacCell(
-            boost::shared_ptr<AbstractIvpOdeSolver> pSolver,
-            boost::shared_ptr<AbstractStimulusFunction> pStimulus)
-    {
-        return new {{class_name}}(pSolver, pStimulus);
-    }
-
-}{%- endif %}
-#endif // CHASTE_CVODE
+{% include "Normal/cpp/header_comments" %}
```
jonc125: I'd be inclined to put things that are shared between Normal and Cvode in a `shared` folder, rather than one using the other - makes it easier to follow the structure, and to see what is common.

chaste_codegen/templates/cvode_model.hpp
```diff
@@ -1,52 +1,10 @@
...
-public:
-    {%- if default_stimulus_equations["equations"]|length > 0 %}
-    boost::shared_ptr<RegularStimulus> UseCellMLDefaultStimulus();
-    {%- endif %}
-    {%- if use_get_intracellular_calcium_concentration %}
-    double GetIntracellularCalciumConcentration();
-    {%- endif %}
+{% include "Normal/hpp/header_comments" %}
+{% include "Normal/hpp/includes" %}
+{% include "Cvode/hpp/class_def" %}
```
jonc125: If this ends up being the only file in `Cvode/hpp` you might want not to have a separate file, but I don't have a strong opinion on that.

PR #111, opened 2020-04-02
Rush larsen

MauriceHendrix:
## Description

- Implements models for Rush Larsen type solver.
- I also added the option for making Rush Larsen and Backwards Euler models to the command line script.
- In the process I had to move most of the linearity checking code out of the Backwards Euler models as I needed to re-use it.
- Also added an optimized RushLarsen model type with partial-evaluation (lookup tables are not implemented)

## Motivation and Context

- It's one of the types implemented in pycml and required in the chaste setup.
- Forgot to add the the backwards euler model to the command line script
- Finally made changed _format_derivative_equations in chaste_model so that I could get it to format my equations

## Types of changes

## Checklist:

## Testing


chaste_codegen/__init__.py
```diff
@@ -6,6 +6,8 @@
 from cellmlmanip.transpiler import Transpiler
 
 from chaste_codegen.be_model import BeModel  # noqa
+from chaste_codegen.rl_model import RlModel  # noqa
```
jonc125, line 9: I'd generally prefer more explicit names even if they are a bit longer, so `BackwardEulerModel` and `RushLarsenModel`.

chaste_codegen/_command_line_script.py
```diff
@@ -14,7 +14,8 @@
 def chaste_codegen():
     # Link names to classes for converting code
     translators = OrderedDict([('Chaste', cg.NormalChasteModel), ('ChasteOpt', cg.OptChasteModel),
-                              ('CVODE', cg.CvodeChasteModel)])
+                              ('CVODE', cg.CvodeChasteModel), ('BckwardsEuler', cg.BeModel),
```
jonc125, line 17: BckwardsEuler -> BackwardsEuler

chaste_codegen/_linearity_check.py
```diff
@@ -0,0 +1,134 @@
...
+    """Returns the state vars whos derivative expressions are nont linear"""
+    # return the state var part from the derivative equations where the rhs is not linear
+    return sorted([eq.lhs.args[0] for eq in derivative_equations
+                   if isinstance(eq.lhs, sp.Derivative) and
+                   eq.lhs.args[0] != membrane_voltage_var and
+                   check_expr(eq.rhs, eq.lhs.args[0], membrane_voltage_var, state_vars) != KINDS.LINEAR],
+                  key=lambda s: printer.doprint(s))
+
+
+def derives_eqs_partial_eval_non_linear(y_derivatives, non_linear_state_vars, membrane_voltage_var, state_vars,
```
jonc125, line 104: This function name doesn't really enlighten me as to what it's doing! Maybe `subst_deriv_eqs_non_linear_vars` or similar?

chaste_codegen/rl_model.py
```diff
@@ -0,0 +1,139 @@
...
+        self._non_linear_state_vars = \
+            get_non_linear_state_vars(self._derivative_equations, self._membrane_voltage_var,
+                                      self._state_vars, self._printer)
+
+        self._derivative_alpha_beta, self._formatted_alpha_beta_eqs, self._vars_in_derivative_alpha_beta = \
+            self._get_formatted_alpha_beta()
+        self._formatted_alpha_beta_eqs = self._format_alpha_beta_eqs()
+
+    def _get_formatted_alpha_beta(self):
+        """Gets the information for rAlphaOrTau, rBetaOrInf and formatted equations"""
```
jonc125, line 27: The names are probably different here - rAlphaOrTau looks like C++ naming convention?

chaste_codegen/rl_model.py
```diff
@@ -0,0 +1,139 @@
...
+                                                                self._get_equations_for)
+
+        for deriv in self._y_derivatives:
+            ab = {'alpha': None}
+            it = {'tau': None}
+            # get match if possible (deiv is linear)
+            if deriv.args[0] not in self._non_linear_state_vars and deriv.args[0] is not self._membrane_voltage_var:
+                eq = [e for e in linear_derivs_eqs if e.lhs == deriv][0]
+                ab = match_alpha_beta(eq.rhs, eq.lhs.args[0])
+                it = match_inf_tau(eq.rhs, eq.lhs.args[0])
```
jonc125, line 62: You could do this only if ab isn't a match?

chaste_codegen/rl_model.py
```diff
@@ -0,0 +1,139 @@
...
+
+            # check if there was a match
+            if ab['alpha'] is not None or it['tau'] is not None:
+                eq = [e for e in linear_derivs_eqs if e.lhs == deriv][0]
+                rAlphaOrTau = ab['alpha'] if ab['alpha'] else it['tau']
```
jonc125, line 67: Might need `if ab['alpha'] is not None` here (and similarly below)

chaste_codegen/rl_model.py
```diff
@@ -0,0 +1,139 @@
...
+                rBetaOrInf = ab['beta'] if ab['beta'] else it['inf']
+                derivative_alpha_beta.append({'type': 'alphabeta' if ab['alpha'] else 'inftau',
+                                              'rAlphaOrTau': self._printer.doprint(rAlphaOrTau),
+                                              'rBetaOrInf': self._printer.doprint(rBetaOrInf)})
+                vars_in_derivative_alpha_beta.update(rAlphaOrTau.free_symbols | rBetaOrInf.free_symbols)
+            else:
+                derivative_alpha_beta.append({'type': 'non_linear', 'deriv': self._printer.doprint(deriv)})
+                vars_in_derivative_alpha_beta.add(deriv)
+
+        deriv_eqs_EvaluateEquations = self._get_equations_for(vars_in_derivative_alpha_beta)
```
jonc125, line 77: Again this looks more like C++ style naming.

chaste_codegen/rl_model.py
```diff
@@ -0,0 +1,139 @@
...
+
+        # Get all used variables for derivative eqs
+        deriv_variables = set()
+        for eq in deriv_eqs_EvaluateEquations:
+            deriv_variables.update(eq.rhs.free_symbols)
+
+        for sv in self._formatted_state_vars:
+            sv['in_ab'] = sv['sympy_var'] in deriv_variables
+        # quicker way?
```
jonc125, line 86: ?

chaste_codegen/rl_model.py
```diff
@@ -0,0 +1,139 @@
...
+    def _format_alpha_beta_eqs(self):
+        return self._format_derivative_equations(self._formatted_alpha_beta_eqs)
+
+    def generate_chaste_code(self):
+        """ Generates and stores chaste code for the Normal model"""
+
+        # Generate hpp for model
+        template = cg.load_template('rl_model.hpp')
+        self.generated_hpp = template.render({
+            'converter_version': cg.__version__,
```
jonc125, line 98: Given quite a few of these fields are common, I wonder if a method or two on `ChasteModel` returning a basic dictionary that subclasses could extend would be useful?

chaste_codegen/rl_model.py
MauriceHendrix: I don't really understand what you mean? This bit is to figure out whether there is a match first. I do need the elaborate if as otherwise the equation doesn't exist. I could possible do the eq check outside minus the index and just check len>0 in the if.

chaste_codegen/rl_model.py
MauriceHendrix: what's wrong with ab['alpha'] ?

chaste_codegen/rl_model.py
MauriceHendrix: uups an old comment I did some elaborate slow way before and knew it wasn't efficient

chaste_codegen/rl_model.py
MauriceHendrix: that's a nice idea actually why didn't I think of that

chaste_codegen/rl_model.py
jonc125: As in, only check for an inf_tau match in this case. But sounds like you figured that out already :)

PR #112, opened 2020-04-04
Rush larsen

MauriceHendrix:
## Description

Few formatting tweaks to RushLarsen & RushLarsenOpt models

## Motivation and Context

Forgot to test the RushLarsen opt models in Chaste, this PR includes a few formatting tweaks to prevent unused variable errors they were giving.

## Types of changes

## Checklist:

## Testing
Also tested in chaste


chaste_codegen/rush_larsen_opt_model.py
```diff
@@ -13,7 +15,25 @@ def _get_extended_equations_for_ionic_vars(self):
...
+                         keep_multiple_usages=False)
+        # Get all used variables for derivative eqs
+        deriv_variables = set()
+        for eq in self._vars_for_template['derivative_alpha_beta_eqs']:
+            deriv_variables.update(eq.rhs.free_symbols)
+
+        for sv in self._formatted_state_vars:
+            sv['in_ab'] = sv['sympy_var'] in deriv_variables
+
+    def _format_alpha_beta_eqs(self):
```
jonc125, line 31: Is there a reason we can't use the base class method here? Why the different logic for `in_eqs_excl_voltage` (which seems to be the only difference)?

MauriceHendrix: I see where you're coming from. 
The thing is that for the self._derivative_eqs_excl_voltage is a subset of the original derivative equations, so not partially evaluated. When you apply the partial eval,the only equation that's not in there is: dx/dt = ... so instead of applying the partial eval I just wrote that. (The rest of the result of partial eval you won't see anyway in these types of models).

 Maybe it would be more logical though to just apply the partial eval?

jonc125: If you get the same result either way, I'd be inclined to avoid the code duplication I think.

jonc125: Or at least add a comment to explain.

MauriceHendrix: Ok makse sense I rewrote it to apply partial eval and use super class formatting rules.

PR #115, opened 2020-04-08
moved jacobian functions and added a few bits to normal model

MauriceHendrix:
## Description

moved jacobian functions and added a few bits to normal model in prep for Generalised Rush Larsen

## Motivation and Context

- Generalised Rush Larsencan use the same jacobian code
- Actually backwards Euler can also use the same jacobian code as CVODE
- For Generalised Rush Larsen it's conveniant to have access to smypy variables in formatted dicts so I can quickly update them

## Types of changes

## Checklist:

## Testing

### DIFF
--- a/chaste_codegen/_jacobian.py
+++ b/chaste_codegen/_jacobian.py
@@ -2,7 +2,14 @@ import sympy as sp
 
 
 def get_jacobian(state_vars, derivative_equations):
-    """Calculate the analytic jacobian"""
+    """Calculate the analytic jacobian
+    
+    :param state_vars: set of state variables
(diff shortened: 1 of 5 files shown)


chaste_codegen/_jacobian.py
```diff
@@ -0,0 +1,33 @@
+import sympy as sp
+
+
+def get_jacobian(state_vars, derivative_equations):
```
jonc125, line 4: Would be good to add to the docstrings for both functions things like what is returned, what the parameters are.

MauriceHendrix: Yeah fair point I'll do that.

PR #123, opened 2020-04-20
fixed bug in expose annotation

MauriceHendrix:
## Description

Bug fix in expose_modifiable_parameters

## Motivation and Context

- While working on data-clamp I had a close look at expose_annotation and noticedthe modifiable parameters were not quite correct, Too many stimulus parameters were being filtered out and once I fixed that it needed some adjustments to make them show up properly
- This reference actually has the correct modifiable parameters as per pycml

## Types of changes

## Checklist:

n/a

## Testing


chaste_codegen/chaste_model.py
```diff
@@ -205,10 +205,10 @@ def __init__(self, model, file_name, **kwargs):
              'derived_quantities': self._format_derived_quant(),
              'derived_quantity_equations': self._format_derived_quant_eqs()}
 
-    def _get_equations_for(self, variables, recurse=True):
+    def _get_equations_for(self, variables, recurse=True, filter_modifiable_parameters_lhs=True):
```
jonc125, line 208: Maybe document function parameters?

Also, if subclasses are intended to use a method, perhaps it shouldn't start with an underscore?

PR #124, opened 2020-04-21
exposed V and cai as modifiable parameters / derived quantities as ap…

MauriceHendrix:
…propriate

## Description

Exposes V and Cai as modifiable parameters, (or derived quanities if they're not constant)

## Motivation and Context

Gary said they should be marked. pycml code also does this when it maks them as input.

However while we can get the modifiable parameters bit in pycml, it doesn't seem to do the derived quantities bit??

## Types of changes

## Checklist:

## Testing

- The modifiable parameters bit can be seen in action in `zhang_SAN_model_2000_0D_capable.cpp` (see \chaste_codegen\data\tests\chaste_reference_models\Normal)
- The derived quantities bit can be found in modifiable parameters bit can be seen in action in `matsuoka_model_2003.cpp` (see \chaste_codegen\data\tests\chaste_reference_models\GRL1Opt)


chaste_codegen/chaste_model.py
```diff
@@ -309,8 +313,13 @@ def _get_cytosolic_calcium_concentration_var(self):
...
 
     def _get_modifiable_parameters_annotated(self):
-        """ Get the variables annotated in the model as modifiable parametery"""
-        return self._model.get_variables_by_rdf((self._PYCMLMETA, 'modifiable-parameter'), 'yes')
+        """ Get the variables annotated in the model as modifiable parameter
+
+            Also include V and cytosolic_calcium_concentration_var if these are constant and not stae_vars
+        """
+        return self._model.get_variables_by_rdf((self._PYCMLMETA, 'modifiable-parameter'), 'yes') +\
+            [p for p in (self._membrane_voltage_var, self._cytosolic_calcium_concentration_var)
```
jonc125, line 321: Better would be to set the modifiable-parameter annotation for these variables in earlier processing.

chaste_codegen/chaste_model.py
```diff
@@ -330,10 +339,10 @@ def _get_modifiable_parameters(self):
         if self.expose_annotated_variables:
             # Combined and sorted in display name order
             return \
-                sorted(self._get_modifiable_parameters_annotated() + self._get_modifiable_parameters_exposed(),
+                sorted(set(self._get_modifiable_parameters_annotated() + self._get_modifiable_parameters_exposed()),
```
jonc125, line 342: Which would avoid needing to cast to & from set here.

chaste_codegen/chaste_model.py
```diff
@@ -656,8 +665,13 @@ def _get_derivative_eqs_voltage(self):
...
 
     def _get_derived_quant_annotated(self):
-        """ Get the variables annotated in the model as derived derived-quantity"""
-        return self._model.get_variables_by_rdf((self._PYCMLMETA, 'derived-quantity'), 'yes')
+        """ Get the variables annotated in the model as derived derived-quantity
+
+            Also include V and cytosolic_calcium_concentration_var if these are not constant and not stae_vars
+        """
+        return self._model.get_variables_by_rdf((self._PYCMLMETA, 'derived-quantity'), 'yes') +\
+            [p for p in (self._membrane_voltage_var, self._cytosolic_calcium_concentration_var)
```
jonc125, line 673: Likewise for these, set the annotation if appropriate.

chaste_codegen/data/tests/all_models_reference_models/BE/matsuoka_model_2003.cpp
```diff
@@ -40,6 +40,7 @@
         return p_cellml_stim;
     }
 
+
```
jonc125, line 43: Avoid the extra blank line?

chaste_codegen/chaste_model.py
MauriceHendrix: ok

chaste_codegen/chaste_model.py
MauriceHendrix: actually no it would still be necessary because cytosolic_calcium_concentration will now show up both through the metadata and through the fact it has annotation. It should have actually always been there realy to capture this case

MauriceHendrix:
> OK. I'd add a comment explaining that then :)

actually I'll add an update to those 2 methods instead bound to be quicker

chaste_codegen/chaste_model.py
```diff
@@ -279,12 +291,22 @@ def _get_membrane_voltage_var(self):
         """ Find the membrane_voltage variable"""
         voltage = self._model.get_variable_by_ontology_term((self._OXMETA, "membrane_voltage"))
         desired_units = self._units.get_unit('millivolt')
+        # If V is not a state var add annotation as modifiable parameter or derived quantity as appropriate.
+        if voltage not in self._state_vars:
```
jonc125, line 295: Could extract the logic here as a method so it can be called for calcium below too?

MauriceHendrix: yeah good point I'll do that

PR #132, opened 2020-04-27
switch to using cellmlmanip model.get_display_name

MauriceHendrix:
## Description

Switched to using model.get_display_name in cellmlmanip instead of our own function

## Motivation and Context

Might as well use the library since we've added it there now

## Types of changes

Switch to using new library call instead of own code

## Checklist:

n/a

## Testing

### DIFF
--- a/chaste_codegen/chaste_model.py
+++ b/chaste_codegen/chaste_model.py
@@ -256,17 +256,6 @@ class ChasteModel(object):
         else:
             return self._MEMBRANE_VOLTAGE_INDEX + self._CYTOSOLIC_CALCIUM_CONCENTRATION_INDEX + 1
 
-    def _get_var_display_name(self, var):
-        """Return a display name for the given variable.
-
-        Looks for OXMETA ontology annotation tage first, then cmeta:id if present, or the name attribute if not.
-        If there is an interface component, strip the name of it out of the display name.
-        """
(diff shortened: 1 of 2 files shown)


chaste_codegen/chaste_model.py
```diff
@@ -256,17 +256,6 @@ def _state_var_key_order(self, var):
...
-    def _get_var_display_name(self, var):
-        """Return a display name for the given variable.
-
-        Looks for OXMETA ontology annotation tage first, then cmeta:id if present, or the name attribute if not.
-        If there is an interface component, strip the name of it out of the display name.
-        """
-        display_name = var.cmeta_id if var.cmeta_id else var.name
-        if self._model.has_ontology_annotation(var, self._OXMETA):
-            display_name = self._model.get_ontology_terms_by_variable(var, self._OXMETA)[-1]
-        return display_name.replace('$', '__').replace('var_chaste_interface_', '', 1).replace('var_', '', 1)
```
jonc125, line 268: Given our `Model` method doesn't do all the Chaste-specific changes, do you want to keep this method but have it call `self.model.get_display_name` for the common parts?

MauriceHendrix: I think we don't need the `replace('var_chaste_interface_', '', 1).replace('var_', '', 1)` anymore
and I think the other replace needs to go in cellmlmanip to be able to sort consitent with pycml?

PR #134, opened 2020-04-28
fixed expose only including stimulus current if explicitly annotated

MauriceHendrix:
## Description

As we decided to make exposing annotation no longer optional, I've removed the option and had to regenerate reference files

## Motivation and Context

The option wasn't quite doing what was needed, and it was agreed most people would be confused by it not being on.

## Types of changes

## Checklist:

## Testing
I will also check the new models in chaste (not yet done).


chaste_codegen/chaste_model.py
```diff
@@ -321,33 +320,19 @@ def _get_cytosolic_calcium_concentration_var(self):
...
-            return \
-                sorted(set(self._get_modifiable_parameters_annotated() + self._get_modifiable_parameters_exposed()),
-                       key=lambda v: self._model.get_display_name(v, self._OXMETA))
-        else:
-            return sorted(self._get_modifiable_parameters_annotated(),
-                          key=lambda v: self._model.get_display_name(v, self._OXMETA))
+        """ Get all modifiable parameters, either annotated as such or with other annotation.
+
+        Stimulus currents are ignored and the result is sorted by display name"""
+        return sorted(set([q for q in list(self._model.variables()) +
```
jonc125, line 327: This is a long comprehension! Would something like the following be more readable?
```python
annotated_vars = set(
    [var for var in self._model.variables() if self._model.has_ontology_annotation(var, self._OXMETA)] +
    self._model.get_variables_by_rdf((self._PYCMLMETA, 'modifiable-parameter'), 'yes'))
current_vars = set(
    [var for var in annotated_vars
     if self._model.get_ontology_terms_by_variable(var, self._OXMETA)[-1].startswith('membrane_stimulus_current')])
parameters = annotated_vars - current_vars - set(self._model.get_derived_quantities() + self._state_vars + [self._time_variable])
return sorted(parameters, key=lambda v: self._model.get_display_name(v, self._OXMETA))
```

chaste_codegen/chaste_model.py
```diff
@@ -664,29 +649,15 @@ def _get_derivative_eqs_voltage(self):
...
-            variables exposed with oxford metadata that are derived quantities but are not annotated as such"""
-        if self.expose_annotated_variables:
-            # Combined and sorted in display name order
-            return \
-                sorted(set(self._get_derived_quant_annotated() + self._get_derived_quant_exposed()),
-                       key=lambda v: self._model.get_display_name(v, self._OXMETA))
-        else:
-            return self._get_derived_quant_annotated()  # These are already sorted
+        Stimulus currents are ignored and the result is sorted by display name"""
+        return sorted(set([q for q in self._model.get_derived_quantities() +
```
jonc125, line 656: This will only keep vars with an explicit derived-quantity annotation if they also have an oxmeta annotation?

chaste_codegen/templates/cvode_with_data_clamp/cpp/ComputeDerivedQuantities
```diff
@@ -8,19 +8,17 @@
         {%- if state_var.in_derived_quant %}double {{ state_var.var }} = NV_Ith_S(rY,{{loop.index0}});
         // Units: {{state_var.units}}; Initial value: {{state_var.initial_value}}
         {% endif %}{%- endfor %}
-
+        {%- for eq in derived_quantity_equations %}{%- if eq.is_data_clamp_current %}
```
jonc125, line 11: This doesn't look right to me! You only want the 'mathematics' comment once.

chaste_codegen/chaste_model.py
MauriceHendrix: O yes uups

MauriceHendrix:
> Did you regenerate the references using the old code with the option on, to check that you're getting the same results once you'd stripped the option out?

good idea I'll do that!

PR #136, opened 2020-05-05
time as derived quantity

MauriceHendrix:
## Description

annotated time as a derived quantity

## Motivation and Context

makse sure it shows up as a derived quantity in generated code, as that seems to be what we agreed should happen

## Types of changes

## Checklist:

## Testing


chaste_codegen/chaste_model.py
```diff
@@ -266,6 +266,10 @@ def _add_units(self):
     def _get_time_variable(self):
         time_variable = self._model.get_free_variable()
         desired_units = self._units.get_unit('millisecond')
+        # Add derived quantity metadata tag
+        self._model.add_cmeta_id(time_variable)
+        self._model.rdf.add((time_variable.rdf_identity, create_rdf_node((self._PYCMLMETA, 'derived-quantity')),
+                            create_rdf_node('yes')))
```
jonc125, line 272: Is it worth adding some methods like `self._set_is_derived_quantity(var)` which do this, and similarly for modifiable-parameter?

PR #144, opened 2020-05-27
Convert other currents (if possible)

MauriceHendrix:
## Description

Converts other currents (besides stimulus current) if possible and if converted adds a derived quantity in chaste units, using the ontology.
Some things that are currents actually aren't quite currents and in those cases we don't have any way to convert them so we skip the conversion.

## Motivation and Context

Gary thought this would be nice & it was easy to do (using the already existing code to get things from the ontology)

## Types of changes

## Checklist:

## Testing


chaste_codegen/chaste_model.py
```diff
@@ -400,6 +402,16 @@ def _convert_membrane_stimulus_current(self):
         return membrane_stimulus_current_converted, \
             (self._membrane_stimulus_current_orig.units, membrane_stimulus_current_converted.units)
 
+    def _convert_other_currents(self):
+        """ Try to convert other currents (apart from stimulus current) to uA_per_cm2 """
```
jonc125, line 406: This looks good. Maybe add to the docstring that only variables in units convertible to Chaste's current units are converted, to allow for 'current-related' variables in the ontology that aren't themselves currents but might be included within the superclass.

chaste_codegen/data/tests/all_models_reference_models/BE/Shannon2004.cpp
```diff
@@ -67,7 +67,7 @@
         this->mParameters[8] = 16.0; // (var_INa__G_INa) [milliS_per_microF]
         this->mParameters[9] = 0.029999999999999999; // (var_IKr__G_IKr_const) [milliS_per_microF]
         this->mParameters[10] = 0.070000000000000007; // (var_IKs__G_Ks_Max) [milliS_per_microF]
-        this->mParameters[11] = 9.0; // (var_INaCa__V_max) [microA_per_microF]
+        this->mParameters[11] = 0; // (var_INaCa__V_max) [microA_per_microF]
```
jonc125, line 70: This change worries me!

chaste_codegen/chaste_model.py
```diff
@@ -400,6 +402,16 @@ def _convert_membrane_stimulus_current(self):
...
+        currents = get_variables_transitively(self._model, (self._OXMETA, 'CellMembraneCurrent'))
+        currents = set(currents) - set(self._stimulus_params + [self._membrane_stimulus_current_converted])
+        for current in currents:
+            try:
+                self._model.convert_variable(current, self.units.get_unit('uA_per_cm2'), DataDirectionFlow.INPUT)
```
jonc125, line 411:
```suggestion
                self._model.convert_variable(current, self.units.get_unit('uA_per_cm2'), DataDirectionFlow.OUTPUT)
```
Since Chaste is just reading currents, not modifying them - they're derived quantities, not parameters.

chaste_codegen/data/tests/all_models_reference_models/BE/Shannon2004.cpp
```diff
@@ -199,21 +200,24 @@
...
         const double var_ICaL__i_CaL_Na_jct = (-var_ICaL__gamma_Nao * mParameters[6] + var_ICaL__gamma_Nai * var_chaste_interface__Na_buffer__Na_jct * exp(var_chaste_interface__cell__V * var_model_parameters__F / (var_model_parameters__R * var_model_parameters__T))) * var_ICaL__Fx_ICaL_jct * var_ICaL__PNa * var_ICaL__temp * var_ICaL_fCa_gate__fCa_jct / (-1.0 + exp(var_chaste_interface__cell__V * var_model_parameters__F / (var_model_parameters__R * var_model_parameters__T))); // microA_per_microF
-        const double var_ICaL__i_CaL = var_ICaL__i_CaL_Ca_SL + var_ICaL__i_CaL_Ca_jct + var_ICaL__i_CaL_K + var_ICaL__i_CaL_Na_SL + var_ICaL__i_CaL_Na_jct; // microA_per_microF
+        const double var_ICaL__i_CaL_converted = (var_ICaL__i_CaL_Ca_SL + var_ICaL__i_CaL_Ca_jct + var_ICaL__i_CaL_K + var_ICaL__i_CaL_Na_SL + var_ICaL__i_CaL_Na_jct) * HeartConfig::Instance()->GetCapacitance(); // uA_per_cm2
```
jonc125, line 203: This arises from using INPUT instead of OUTPUT. Should have `const double var_ICaL__i_CaL_converted = var_ICaL__i_CaL * HeartConfig::Instance()->GetCapacitance();`

chaste_codegen/data/tests/all_models_reference_models/BE/Shannon2004.cpp
```diff
@@ -339,6 +347,7 @@
         const double var_INaCa__Ka_SL = 1 / (1.0 + pow(var_INaCa__Kd_act, 3) / pow(var_chaste_interface__Ca_buffer__Ca_SL, 3)); // dimensionless
         const double var_INaCa__Ka_jct = 1 / (1.0 + pow(var_INaCa__Kd_act, 3) / pow(var_chaste_interface__Ca_buffer__Ca_jct, 3)); // dimensionless
         const double var_INaCa__Q10_NCX = 1.5700000000000001; // dimensionless
+        const double var_INaCa__V_max_converted = 9.0 * HeartConfig::Instance()->GetCapacitance(); // uA_per_cm2
```
jonc125, line 350: This is an interesting one!

chaste_codegen/data/tests/all_models_reference_models/BE/Shannon2004.cpp
```diff
@@ -2029,7 +2047,7 @@ void OdeSystemInformation<CellShannon2004FromCellMLBackwardEuler>::Initialise(vo
     this->mParameterUnits.push_back("milliS_per_microF");
 
     // mParameters[11]:
-    this->mParameterNames.push_back("membrane_sodium_calcium_exchanger_current_conductance");
+    this->mParameterNames.push_back("INaCa__V_max");
```
jonc125, line 2050: This change looks wrong too. Hopefully a consequence of the INPUT instead of OUTPUT.

chaste_codegen/data/tests/all_models_reference_models/BE/Shannon2004.cpp
jonc125: It's always worth sanity-checking the diff for a few generated files to see if things like this crop up!

MauriceHendrix: hhm yeah I checked most but not all of them, quite a lot of fukes to check so thanks for spotting it

chaste_codegen/data/tests/all_models_reference_models/BE/Shannon2004.cpp
MauriceHendrix: yes that solves this

chaste_codegen/data/tests/all_models_reference_models/BE/Shannon2004.cpp
MauriceHendrix: also gone with the input/output change

chaste_codegen/data/tests/all_models_reference_models/BE/Shannon2004.cpp
MauriceHendrix: This is a logical consequence of the conversion: the conversion moving the metadata tag so changing to output doesn't change that.

Also: wouldn't it be weird to have a parameter and a derived quantity named the same (but in different units?)

mirams:
There are some `membrane...current_conductance` here being converted to uA_per_cm2, which is not what we want. In fact I'm not sure if should be possible to get from conductance units to these units.

chaste_codegen/data/tests/all_models_reference_models/BE/Shannon2004.cpp
jonc125: Ah, yes, that would be odd! I didn't realise it was both. As Gary says, we don't really want conductances to be considered derived quantities and parameters. Perhaps the logic here needs to be the same as for `GetIIonic` rather than using the ontology? Or we need to fix the ontology to support this use case properly!

jonc125:
At least one of the V_max values is in microA_per_microF. Which is convertible to current units...

MauriceHendrix:
> current_conductance
aah like `membrane_sodium_calcium_exchanger_current_conductance` in `stewart_zhang_model_2008_ss`

But those are currents according to the ontology

chaste_codegen/data/tests/all_models_reference_models/BE/Shannon2004.cpp
MauriceHendrix: This issue should now be fixed due to the ontology fix, this seems to be the case looking at  `chaste_codegen/data/tests/all_models_reference_models/BE/Shannon2004.cpp `

chaste_codegen/data/tests/all_models_reference_models/BE/Shannon2004.cpp
MauriceHendrix: should be fixed due to your suggested direction change

MauriceHendrix:
Using the latest version of the onology I get things that are MembraneCurrenrRelated AND IonCurrents but not stimulus current

chaste_codegen/chaste_model.py
```diff
@@ -400,6 +406,24 @@ def _convert_membrane_stimulus_current(self):
         return membrane_stimulus_current_converted, \
             (self._membrane_stimulus_current_orig.units, membrane_stimulus_current_converted.units)
 
+    def _convert_other_currents(self):
+        """ Try to convert other currents (apart from stimulus current) to uA_per_cm2
+
+        Currents are terms tagged with a term that is both `CellMembraneCurrentRelated` and an `IonicCurrent`
+        according to the OXMETA ontology (https://chaste.comlab.ox.ac.uk/cellml/ns/oxford-metadata#).
+        Only variables in units convertible to Chaste's current units are converted,
+        so to currents in unusual units (e.g. dimensionless) are skipped.
```
jonc125, line 415:
```suggestion
        so currents in unusual units (e.g. dimensionless) are skipped.
```

PR #158, opened 2020-07-10
multi model & options for chaste integartion

MauriceHendrix:
## Description

- Allows multi model processing in command line
- Model only loaded once when generating multi-model types to save time. To this end the conversion and some of the equation processing has moved out of the generator model types, so that models with unit conversions applied can be reused.
- Parameters closer to pycml ones for convenience while integration
- Class names especially for opt reference models changed
- Disallowed multi model generation when -o or --dynamically-loadable options are used
- Fixed modifiers bug which was causing TestCvodeCells (in chaste) to fail
- Fixed sign issue on ionic currents
- Removed `WARNING:rdflib.term: .ttl does not look like a valid URI, trying to serialize this will break.`

## Motivation and Context

- Needed for chaste integration
- Saves time
- Conveniant
- More in line with what pycml generates
- Wouldn't know exactly what files to generate with -o and multi model, and for dynamic build/load it's useful to generate just 1 model, so that it's clear which files are being dynamically loaded (we can just use the existing infrastructure as pycml worked this way too)
- Modifiers weren't working. The reason for this was that get_equations_for filters out lhs if the lhs is a parameter to prevent printing e.g. `mParameter[0] = 25`. Instead we had adjusted the printer to print the parameter as mParameter[0] directly in the equations. This however means we were missing out on modifier definition equations for some modifier. I have adjusted the printing to print variables that are modifiers as modifiers in the equation if they don't have a defining equation. The reason we can't just keep the parameter defining equations is that the jacobian analysys and the rush-larsen pattrn matching want each derivative equation folded into 1 eq, so we can't then keep the parameter defining eq.

The code below illustartes the change:
*pycml version*
```
const double var_INa__G_INa = mp_membrane_fast_sodium_current_conductance_modifier->Calc(mParameters[8], var_chaste_interface__environment__time);
...
const double var_INa__i_Na_jct = 0.11 * var_INa__G_INa * var_INa__openProb * (var_chaste_interface__cell__V - var_reversal_potentials__E_Na_jct);
```
*previous codegen version*
```
const double var_INa__i_Na_jct = (-var_reversal_potentials__E_Na_jct + var_chaste_interface__cell__V) * var_INa__Fx_Na_jct * mParameters[8] * var_INa__openProb; // microA_per_microF
```
*new codegen version*
```
const double var_INa__i_Na_jct = (-var_reversal_potentials__E_Na_jct + var_chaste_interface__cell__V) * var_INa__Fx_Na_jct * mp_membrane_fast_sodium_current_conductance_modifier->Calc(mParameters[8], var_chaste_interface__environment__time) * var_INa__openProb;
```

- I was missing the sign analysis of the dvdt equation that pycml has. But there was also an issue with unit conversion. E.g. for FaberRudy2000 the getiionic result for codegen was - the pycml result. The reason was the conversion of time, whicg turned 
'dvdt = -1.0 * sum of terms` into:
```
dvdt_orig = -1.0 * sum of terms
dvdt=dvdt_orig
```
due to the way the analysis worked it picked up just dvdt_orig as current hence the negation of the result.
The solution used is to find the currents before any conversions take place do the sign analysis, tag them with metadata and pull them out via metadata post conversion, so that we generate the correct (post conversion) equations, but also have the correct sign.

- rdflib warning disappears when giving the parse method a filename rather than a stream

## Types of changes

## Documentation checklist

- [X] I have updated all documentation in the code where necessary.
- [X] I have checked spelling in all (new) comments and documentation.
- [X] I have added a note to RELEASE.md if relevant (new feature, breaking change, or notable bug fix).

## Testing


RELEASE.md
```diff
@@ -1,2 +1,9 @@
+# Release 0.2.0
+- The command line interface now allows generating multiple model types in one go.
+- The command line interface now has a --show-outputs option
+- Some command line argument names have changed to more closely match what pycml used to use, command line arguments are not backwards compatible with release 0.1.0
+- An issue with generating modifiers that are also parameters has been fixed, by generating the modifier in situ if it doesn't have a difining equation.
```
jonc125, line 5:
```suggestion
- An issue with generating modifiers that are also parameters has been fixed, by generating the modifier in situ if it doesn't have a defining equation.
```

RELEASE.md
```diff
@@ -1,2 +1,9 @@
...
+- An issue with getiionic where the sign of the equations was incorrect has been fixed by anlysing the equation before any unit conversion has taken place.
```
jonc125, line 6:
```suggestion
- An issue with GetIIonic where the sign of the equations was incorrect has been fixed by analysing the equation before any unit conversion has taken place.
```

chaste_codegen/_command_line_script.py
```diff
@@ -5,98 +5,161 @@
...
+     ('grl1', (cg.GeneralisedRushLarsenFirstOrderModel, 'FromCellMLGRL1', 'GRL1', False)),
+     ('grl2', (cg.GeneralisedRushLarsenSecondOrderModel, 'FromCellMLGRL2', 'GRL2', False))])
 
-TRANSLATORS_WITH_MODIFIERS = ('Chaste', 'ChasteOpt', 'CVODE', 'CVODEWithDataClamp')
+TRANSLATORS_OPT = OrderedDict(
+    [('normal', (cg.OptChasteModel, 'FromCellMLOpt', 'Opt', True)),
+     ('cvode', (cg.OptCvodeChasteModel, 'FromCellMLCvodeOpt', 'CvodeOpt', True)),
+     ('rush-larsen', (cg.RushLarsenOptModel, 'FromCellMLRushLarsen', 'RushLarsen', False)),
+     ('grl1', (cg.GeneralisedRushLarsenFirstOrderModelOpt, 'FromCellMLGRL1', 'GRL1', False)),
+     ('grl2', (cg.GeneralisedRushLarsenSecondOrderModelOpt, 'FromCellMLGRL2', 'GRL2', False))])
```
jonc125, line 28: Should these have 'Opt' appended to the class name?

chaste_codegen/_command_line_script.py
```diff
@@ -5,98 +5,161 @@
...
-                           ('CVODE', cg.CvodeChasteModel),
-                           ('CVODEWithDataClamp', cg.CvodeWithDataClampModel),
-                           ('BackwardsEuler', cg.BackwardEulerModel),
-                           ('RushLarsen', cg.RushLarsenModel),
-                           ('RushLarsenOpt', cg.RushLarsenOptModel),
-                           ('GeneralisedRushLarsen1', cg.GeneralisedRushLarsenFirstOrderModel),
-                           ('GeneralisedRushLarsen1Opt', cg.GeneralisedRushLarsenFirstOrderModelOpt),
-                           ('GeneralisedRushLarsen2', cg.GeneralisedRushLarsenSecondOrderModel),
-                           ('GeneralisedRushLarsen2Opt', cg.GeneralisedRushLarsenSecondOrderModelOpt)])
+TRANSLATORS = OrderedDict(
```
jonc125, line 14: I'd add a comment about what the fields in the tuples represent!

chaste_codegen/_command_line_script.py
```diff
@@ -5,98 +5,161 @@
...
-                        help='the type of code to output [default: Chaste].  '
-                        'Choices: ' + str(list(TRANSLATORS.keys())))
-    parser.add_argument('-o', dest='outfile', metavar='OUTFILE', default=None,
-                        help='write program code to OUTFILE '
-                             '[default action is to use the input filename with a different extension] '
-                             'NOTE: expects provided OUTFILE to have an extension relevant to code being generated '
-                             '(e.g. for CHASTE/C++ code: .cpp, .c, .hpp, or .h)')
+
+    group = parser.add_argument_group('ModelTypes', 'The different types of solver models for which code that can be '
+                                      'generated, if no model type is set, normal models are generated')
```
jonc125, line 51:
```suggestion
    group = parser.add_argument_group('ModelTypes', 'The different types of solver approach for which code can be '
                                      'generated; if no model type is set, "normal" models are generated')

```

chaste_codegen/_command_line_script.py
```diff
@@ -5,98 +5,161 @@
...
     group = parser.add_argument_group('Generated code options')
-    group.add_argument('-c', default=None, dest='class_name',
+    group.add_argument('-o', dest='outfile', metavar='OUTFILE', default=None,
+                       help='write program code to OUTFILE '
+                            '[default action is to use the input filename with a different extension] '
+                            'NOTE: expects provided OUTFILE to have an extension relevant to code being generated '
+                            '(e.g. for CHASTE/C++ code: .cpp, .c, .hpp, or .h)')
+    group.add_argument('--output-dir', action='store', help="directory to place output files in", default=None)
+    group.add_argument('--show-outputs', action='store_true', default=False,
+                       help="don't actually run PyCml, just show what files would be generated, one per line")
```
jonc125, line 70:
```suggestion
                       help="don't actually generate code, just show what files would be generated, one per line")

```

chaste_codegen/_command_line_script.py
```diff
@@ -5,98 +5,161 @@
...
-    group.add_argument('--use-modifiers', dest='use_modifiers',
+    group.add_argument('--opt', action='store_true', default=False,
+                       help="apply default optimisations to all generated code")
+    group.add_argument('-m', '--use-modifiers', dest='modifiers',
                        action='store_true', default=False,
                        help='add modifier functions for metadata-annotated variables (except time & stimulus) '
-                       'for use in sensitivity analysis. Only works with one of the following translator types (-t): ' +
-                       str(TRANSLATORS_WITH_MODIFIERS))
+                       'for use in sensitivity analysis. Only works with one of the following model types: ' +
+                       str(TRANSLATORS_WITH_MODIFIERS) + ' and is ignored for others')
```
jonc125, line 86:
```suggestion
                       'for use in sensitivity analysis. Only works with one of the following model types and is ignored for others: ' +
                       str(TRANSLATORS_WITH_MODIFIERS))

```

chaste_codegen/_command_line_script.py
```diff
@@ -5,98 +5,161 @@
...
+            # if -o or dynamically_loadable is selected with opt, only convert opt model type
+            if not args.opt or (not args.outfile and not args.dynamically_loadable):
+                translators.append(TRANSLATORS[model_type])
+            if args.opt and model_type in TRANSLATORS_OPT:
+                translators.append(TRANSLATORS_OPT[model_type])
+
+    # An outfile cannot be set with multiple translations
+    if args.outfile and len(translators) > 1:
+        raise ValueError("-o cannot be used when multiple model types have been selected!")
+    # Dynamicly loadable models can only be built one at a time
```
jonc125, line 112:
```suggestion
    # Dynamically loadable models can only be built one at a time

```

chaste_codegen/_command_line_script.py
```diff
@@ -5,98 +5,161 @@
...
+    if args.dynamically_loadable and len(translators) > 1:
+        raise ValueError("Only one output type may be specified if creating a dynamic library!")
```
jonc125, line 114:
```suggestion
        raise ValueError("Only one model type may be specified if creating a dynamic library!")

```

chaste_codegen/chaste_model.py
```diff
@@ -827,3 +416,7 @@ def generate_chaste_code(self):
         # Generate cpp for model
         template = cg.load_template(self._cpp_template)
         self.generated_cpp = template.render(self._vars_for_template)
+
+    def __exit__(self, type, value, traceback):
```
jonc125, line 420: Does this get overridden in subclasses? Because it looks pointless here!

chaste_codegen/_command_line_script.py
```diff
@@ -5,98 +5,161 @@
...
+
+    if not args.show_outputs:
+        model = load_model_with_conversions(args.cellml_file, quiet=args.quiet)  # don't load if only showing output
```
jonc125, line 117:
```suggestion
        # Load model once, not once per translator, but only if we're actually generating code
        model = load_model_with_conversions(args.cellml_file, quiet=args.quiet)
```

chaste_codegen/cvode_opt_chaste_model.py
```diff
@@ -0,0 +1,20 @@
+from chaste_codegen._partial_eval import partial_eval
+from chaste_codegen.cvode_chaste_model import CvodeChasteModel
+
+
+class OptCvodeChasteModel(CvodeChasteModel):
+    """ Holds information specific for the Optimised model type. Builds on Normal model type"""
```
jonc125, line 6: Outdated docstring.

chaste_codegen/data/tests/cellml/test_V_not_state_mparam.cellml
```diff
@@ -44,7 +44,7 @@
         </math>
     </component>
     <component name="membrane">
-        <variable name="time" public_interface="in" units="second"/>
```
jonc125, line 47: Why this change?

chaste_codegen/model_with_conversions.py
```diff
@@ -0,0 +1,502 @@
...
+    simplify,
+    sympify,
+)
+from sympy.codegen.cfunctions import log10
+from sympy.codegen.rewriting import ReplaceOptim, log1p_opt, optimize
+
+from chaste_codegen import LOGGER
+from chaste_codegen._rdf import OXMETA, PYCMLMETA, get_variables_transitively
+
+from ._math_functions import MATH_FUNC_SYMPY_MAPPING
```
jonc125, line 26: Why two different styles of relative import for chaste_codegen?

chaste_codegen/model_with_conversions.py
```diff
@@ -0,0 +1,502 @@
...
+    lambda e: (  # cost function prevents turning log(x) into log(10) * log10(x) as we want normal log in that case
+        e.is_Pow and e.exp.is_negative  # division
+        or (isinstance(e, (log, log10)) and not e.args[0].is_number))))
+# For P^n make sure n is passed as int if it is actually a whole number
+_POW_OPT = ReplaceOptim(lambda p: p.is_Pow and (isinstance(p.exp, Float) or isinstance(p.exp, float))
+                        and float(p.exp).is_integer(),
+                        lambda p: Pow(p.base, int(float(p.exp))))
+
+
+def load_model_with_conversions(model_file, use_modifiers=True, quiet=False):
```
jonc125, line 52: I would have thought `False` was a better default for `use_modifiers`.

chaste_codegen/model_with_conversions.py
```diff
@@ -0,0 +1,502 @@
...
+    if quiet:
+        LOGGER.setLevel(logging.ERROR)
+    model = cellmlmanip.load_model(model_file)
+    add_conversions(model, use_modifiers)
+    return model
+
+
+def add_conversions(model, use_modifiers=True):
+    # Add units needed for conversions
+    model.conversion_units, model.stimulus_units = _add_units(model)
```
jonc125, line 62: You're adding attributes to the cellmlmanip `Model` which has the potential to break if that library adds new fields. Relatively low risk, but maybe worth a comment?

chaste_codegen/model_with_conversions.py
```diff
@@ -0,0 +1,502 @@
...
+    """ If it is not a state var, annotates var as modifiable parameter or derived quantity as appropriate"""
+    if var not in model.state_vars:
+        if model.is_constant(var):
+            set_is_metadata(model, var, 'modifiable-parameter')
+        else:  # not constant
+            set_is_metadata(model, var, 'derived-quantity')
+
+
+def get_equations_for(model, variables, recurse=True, filter_modifiable_parameters_lhs=True, optimise=True):
+    """Returns equations excluding once where lhs is a modifiable parameter
```
jonc125, line 121:
```suggestion
    """Returns equations excluding ones where the lhs is a modifiable parameter
```

chaste_codegen/model_with_conversions.py
```diff
@@ -0,0 +1,502 @@
...
+
+    :param variables: the variables to get defining equations for.
+    :param recurse: recurse and get defining equations for all variables in the defining equations?
+    :param filter_modifiable_parameters_lhs: remove equations where the lhs is a modifiable paramater?
+    :return: List of equations defining vars,
+            with optimisations around using log10, and powers of whole numbers applied to rhs
+            as well as modifiable parameters filtered out is required.
```
jonc125, line 128:
```suggestion
            as well as modifiable parameters filtered out if required.
```

chaste_codegen/tests/test_linearity_check.py
```diff
@@ -38,6 +32,10 @@ def membrane_voltage_var(hn_model):
 
 @pytest.fixture(scope='session')
 def non_linear_state_vars(derivatives_eqs, membrane_voltage_var, state_vars):
+    print(derivatives_eqs)
+    print(membrane_voltage_var)
+    print(state_vars)
+    print('--')
```
jonc125, line 38:
```suggestion
```

chaste_codegen/chaste_model.py
MauriceHendrix: yes but only in `CvodeWithDataClampModel` the reason it's there is that I added the data clamp to the cellmlmanip model, so have to remove it again for the model object to be reusable. I added this to the base class more as a signature and reminder in case we add more types in future really.

chaste_codegen/model_with_conversions.py
MauriceHendrix: thanks will fix for consitency

chaste_codegen/data/tests/cellml/test_V_not_state_mparam.cellml
MauriceHendrix: when I rant these files through pycml it complained that this couldn't be public interface in, but was happy when I left it alltogether, also seemed to work for codegen

chaste_codegen/model_with_conversions.py
MauriceHendrix: good point

chaste_codegen/_command_line_script.py
MauriceHendrix: yes good idea thanks!

chaste_codegen/_command_line_script.py
MauriceHendrix: Would seem to make sense although pycml didn't seem to do this for those models I'll try with opt and see if I can still pass chaste tests

chaste_codegen/chaste_model.py
jonc125: I would add to the docstring for `__enter__` and `__exit__` that they are defined simply here but may be overridden in sub-classes if state needs to be reset for subsequent code generation.

MauriceHendrix: Ok I've also made a note of that these are required wo be able to use them in context.
The way the cleanup for the data clamp model works ia that the model is used  in context in the console script. I tried doing it in a destructor, but I couldn't be sure of exactly when this would be called and it's quite important it happens before subsequent model generations

```
with translator_class(model, outfile_base, header_ext=ext[0], **vars(args)) as chaste_model:
                chaste_model.generate_chaste_code()
```

chaste_codegen/model_with_conversions.py
MauriceHendrix: Good point. I was thinking about this when I did it. I have added an assert (+ comments) to check the model doesn't already hve the attributes we are going to add, that way we'll get a more helpful error message if this ever becomes a problem.

PR #165, opened 2020-08-13
Sqrt

MauriceHendrix:
## Description

Print pow(x, 0.5) as sqrt(x)

## Motivation and Context

- Following Michael's comment after my question on nth squares I checked and it turns out the sympy equations contain pow(x, 0.5) which is what we were printing. I updated the printer to print sqrt(x) as it seems neater and is faster in c++ apparently

## Types of changes

## Documentation checklist

- [ ] I have checked spelling in all (new) comments and documentation.
- [x] I have added a note to RELEASE.md if relevant (new feature, breaking change, or notable bug fix).

## Testing
Will test it in chaste before merging


RELEASE.md
```diff
@@ -1,3 +1,6 @@
+# Release 0.4.1
+- Now outputs sqrt(x) instead of pow(x, 0.5)
```
jonc125, line 2:
```suggestion
- Now outputs sqrt(x) instead of pow(x, 0.5).
```

chaste_codegen/data/tests/all_models_reference_models/BE/FaberRudy2000.cpp
```diff
@@ -1,6 +1,6 @@
 //! @file
 //!
-//! This source file was generated from CellML by chaste_codegen version 0.4.0
+//! This source file was generated from CellML by chaste_codegen version 0.4.1
```
jonc125, line 3: We've discussed before hiding changes from commits where the only difference is a version number like this. Worth considering here?

chaste_codegen/data/tests/all_models_reference_models/BE/nygren_atrial_model_1998.cpp
```diff
@@ -463,7 +463,7 @@
         const double var_x84 = var_x59 * var_x83;
         const double var_x85 = var_x72 * var_x83;
         const double var_x86 = var_x79 * pow(var_chaste_interface__intracellular_ion_concentrations__Na_i, 2.0) * var_chaste_interface__cleft_space_ion_concentrations__K_c / pow((1 + 0.027410122234342148 * var_x74), 2);
-        const double var_x87 = var_x80 * pow(var_chaste_interface__intracellular_ion_concentrations__Na_i, 0.5) * var_chaste_interface__cleft_space_ion_concentrations__K_c;
+        const double var_x87 = var_x80 * sqrt(var_chaste_interface__intracellular_ion_concentrations__Na_i) * var_chaste_interface__cleft_space_ion_concentrations__K_c;
```
jonc125, line 466: Managed to find an example that did change, good!

chaste_codegen/data/tests/all_models_reference_models/BE/FaberRudy2000.cpp
MauriceHendrix: I did apply your suggestions back than but it doesn't seem to be working, although it does seem to work for the date...

MauriceHendrix: aah found it!

PR #176, opened 2020-11-06
better support for secondary trig functions

MauriceHendrix:
## Description

Better support for secondry trig functions

## Motivation and Context

Although not really used in the existing models tested callmlmanip has gained support for these functions so we might as well support them too.

## Types of changes

## Documentation checklist

- [X] I have updated all documentation in the code where necessary.
- [ ] I have checked spelling in all (new) comments and documentation.

## Testing

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,3 +1,6 @@
+# (unreleased)
+- Improved support for secondary trigonometric functions such as sec and acoth.
+
 # Release 0.5.3
 - Added an additional error messages if cellml files can't be loaded and a warning if a lookup table is specified for a tag not present in the model.
 
--- a/chaste_codegen/_chaste_printer.py
+++ b/chaste_codegen/_chaste_printer.py
@@ -65,6 +65,22 @@ class ChastePrinter(Printer):
         'HeartConfig::Instance()->GetCapacitance': 'HeartConfig::Instance()->GetCapacitance',
         'GetExperimentalVoltageAtTimeT': 'GetExperimentalVoltageAtTimeT'
     }
+    _extra_trig_names = {
+        'sec': 'cos',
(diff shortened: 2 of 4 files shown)


MichaelClerx (approved):

mirams (commented):
Don't quite get these, it seems to be saying "sec is another name for cos" etc., but sec = 1/cos ?

MichaelClerx:
The code that handles them is in cellmlmanip https://github.com/ModellingWebLab/cellmlmanip/blob/master/cellmlmanip/printer.py#L213-L237

MichaelClerx:
Just noticed that I didn't add tests (or think about) brackets for the 1/x cases - could go wrong if the rules for functions are being used?

mirams:
OK, maybe nicer to rename it from `extra names` to `reciprocals`?

MauriceHendrix:
> OK, maybe nicer to rename it from `extra names` to `reciprocals`? Or such like

Actually that's up to cellmlmanip since it's inherited I can't go changing those names :)

mirams:
> Just noticed that I didn't add tests (or think about) brackets for the 1/x cases - could go wrong if the rules for functions are being used?

Yeah, worth sticking some extra brackets and tests in there!

MauriceHendrix:
> Just noticed that I didn't add tests (or think about) brackets for the 1/x cases - could go wrong if the rules for functions are being used?

Following C++ code 1st one gets from 1/sec(x), 2nd one I've stuck brackets in
```
float x = 12.12;
std::cout << 1 / 1 / cos(x)<<std::endl;
std::cout << 1 / (1 / cos(x))<<std::endl;

```

```
1.10862
0.90202
```

sorry for not noticing this when reviewing originally

PR #184, opened 2021-01-12
moved optimise and improved used var analysis

MauriceHendrix:
## Description

- Moved expression optimisation for C++ (with regards to pow & log) into its own module so it can be more easily used by singularity analysis code
- changed to that anlyses wich variables are iin use as I realised I was doing an unneeded lop in backward_euler_model.py 
-  and chaste_model.py

## Motivation and Context

- preparation for singularity analysis code (making a seperate PR to reduce reviewing burden)

## Types of changes

## Documentation checklist

- [x] I have checked spelling in all (new) comments and documentation.

This particular change changes nothing in the functionality /api so no need to write in release.md

## Testing

### DIFF
--- a/chaste_codegen/_optimize.py
+++ b/chaste_codegen/_optimize.py
@@ -21,7 +21,7 @@ _POW_OPT = ReplaceOptim(lambda p: p.is_Pow and (isinstance(p.exp, Float) or isin
                         lambda p: Pow(p.base, int(float(p.exp))))
 
 
-def optimize_expr(expr):
+def optimize_expr_for_c_output(expr):
     """Returns expression optimised for c++ export with regards to powers and logarithms
 
     :param expr: the expression to apply power and log optimisations to.
--- a/chaste_codegen/_optimize.py
+++ b/chaste_codegen/_optimize.py
@@ -24,7 +24,7 @@ _POW_OPT = ReplaceOptim(lambda p: p.is_Pow and (isinstance(p.exp, Float) or isin
 def optimize_expr(expr):
     """Returns expression optimised for c++ export with regards to powers and logarithms
 
-    :param expr: the expression to apply power and log optimisations to to get defining equations for.
+    :param expr: the expression to apply power and log optimisations to.
     :return: The expression with the optimisations applied.
     """
     optims = tuple()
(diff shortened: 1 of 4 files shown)


chaste_codegen/_optimize.py
```diff
@@ -0,0 +1,38 @@
...
+        e.is_Pow and e.exp.is_negative  # division
+        or (isinstance(e, (log, log10)) and not e.args[0].is_number))))
+# For P^n make sure n is passed as int if it is actually a whole number
+_POW_OPT = ReplaceOptim(lambda p: p.is_Pow and (isinstance(p.exp, Float) or isinstance(p.exp, float))
+                        and float(p.exp).is_integer(),
+                        lambda p: Pow(p.base, int(float(p.exp))))
+
+
+def optimize_expr(expr):
+    """Returns expression optimised for c++ export with regards to powers and logarithms
```
MichaelClerx, line 25: Could consider a different word than "optimise" because that could mean lots of things (just a suggestion)

chaste_codegen/_optimize.py
```diff
@@ -0,0 +1,38 @@
...
+
+    :param expr: the expression to apply power and log optimisations to to get defining equations for.
```
MichaelClerx, line 27: remove "to get defining equations for" ?

chaste_codegen/model_with_conversions.py
```diff
@@ -11,17 +11,13 @@
...
-    log,
     simplify,
     sympify,
 )
-from sympy.codegen.cfunctions import log10
-from sympy.codegen.rewriting import ReplaceOptim, log1p_opt, optimize
 
 from chaste_codegen import LOGGER, CodegenError
 from chaste_codegen._math_functions import MATH_FUNC_SYMPY_MAPPING
+from chaste_codegen._optimize import optimize_expr
```
MichaelClerx, line 20: It might be good to rethink the organisation a bit?

Usually the `_hidden_modules` are only used when their methods are imported by some _library_, which then exposes them as part of the public API. Chaste codegen seems to be an application rather than a library, so maybe this isn't the best way to do it?

MichaelClerx (approved):
Looks good to me. Wrote down some thoughts.

PR #289, opened 2025-10-29
Add DS_Store entries

mirams:
## Description
Just add DS_Store entries to the gitignore file

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -114,5 +114,8 @@ ENV/
 # Chaste_cg specific
 chaste_codegen/data/tests/chaste_reference_models/develop/
 
+#Mac
+**DS_Store
+
 #windows
 desktop.ini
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -114,5 +114,8 @@ ENV/
 # Chaste_cg specific
 chaste_codegen/data/tests/chaste_reference_models/develop/
 
+#Mac
+**DS_Store
```
kwabenantim:
```suggestion
.DS_Store
```

`**DS_Store` is probably fine. It could potentially match files we didn't intend, though.

kwabenantim: Are you okay for me to change it @mirams?

mirams: Maybe **.DS_Store ? (I think you need the ** for 'within any subdirectory' ?)

kwabenantim: I think using just `.DS_Store` also covers `**/.DS_Store` ([https://stackoverflow.com/questions/107701](https://stackoverflow.com/questions/107701/how-can-i-remove-ds-store-files-from-a-git-repository))

kwabenantim (approved):

